PR #165, opened 2026-05-08
feat: open images via Finder "Open With" → straight into the annotate editor

### LINKED ISSUE #164: [Feature]: Add "Snapzy" to the macOS "Open With" right-click menu
### Problem or Motivation

I'm frustrated when I want to quickly edit an existing image that wasn't captured by Snapzy (e.g. a screenshot from iPhone, an image from Slack/AirDrop, or any PNG/JPG already on disk).

Right now Snapzy doesn't appear in Finder's right-click **Open With** menu, so I have to either launch the app first and drag the file in, or pick "Other…" and locate Snapzy manually every time. This breaks the flow for what should be a one-click action.

Image

### Proposed Solution

It would be great if Snapzy could register itself as a handler for common image types (PNG, JPG, HEIC, GIF, TIFF, WebP) so that:

- Right-click an image in Finder → **Open With → Snapzy** works out of the box.
- Choosing it opens the image **directly in Snapzy's editor**, ready for annotation/cropping — same state as if I had dropped it onto the window.

On macOS this is typically done via `CFBundleDocumentTypes` + `LSItemContentTypes` in `Info.plist`, with `LSHandlerRank = Alternate` so Snapzy doesn't override Preview as the default unless the user opts in.

### Acceptance criteria
- [ ] Snapzy shows up in Finder's **Open With** submenu for PNG/JPG/HEIC/GIF.
- [ ] Selecting it opens the file straight into the edit view.
- [ ] Multi-select opens each image into the editor (or at least the first one).

Thanks! 🙌

### Alternatives Considered

_No response_

### Feature Area

Screenshots

### Mockups / References

_No response_

### Additional Context

_No response_

vnixx:
Closes #164

## Summary

Snapzy now appears in Finder's right-click **Open With** submenu for common image formats (PNG, JPEG, HEIC, HEIF, TIFF, GIF, WebP, BMP). Selecting a file from there — or double-clicking a file whose default app has been set to Snapzy, or dropping image files on the Dock icon — opens it **straight into the existing annotation editor**, ready for cropping / drawing / text, with the same state as if the file had been dropped into the window.

This addresses the friction described in #164: when working with screenshots that weren't taken via Snapzy itself (iPhone screenshots via AirDrop, images from Slack, files saved to disk, etc.), users currently have to launch the app first and drag-drop, or pick `Other…` every time. After this change, the natural macOS gesture just works.

## Changes

Only **2 files**, +73 lines, no refactors.

### 1. `Snapzy/Resources/Info.plist`

Adds a single `CFBundleDocumentTypes` entry declaring Snapzy as an **editor** for the 8 image UTIs above, with:

\`\`\`xml
<key>LSHandlerRank</key>
<string>Alternate</string>
\`\`\`

\`Alternate\` is intentional — Snapzy will surface in the Open With submenu, but **won't override Preview as the system-wide default image viewer** unless the user explicitly chooses to. This avoids any disruptive change for existing users.

### 2. `Snapzy/App/SnapzyApp.swift`

Adds an \`application(_:open:)\` hook on \`AppDelegate\` that:

- Filters incoming URLs down to file URLs only (\`urls.filter { \$0.isFileURL }\`)
- Routes each one to the **existing** \`AnnotateManager.shared.openAnnotation(url:)\` entry point — the same one already used after a screenshot is captured. No new business logic; reuses the well-tested path that already handles sandboxed file access (\`SandboxFileAccessManager.beginAccessingURL\`) and Retina scaling.
- Queues files arriving before \`applicationDidFinishLaunching\` (cold launch via Open With) and flushes once the coordinator is ready, mirroring the existing \`pendingDeepLinkURLs\` pattern in the same file.

## Design notes

**Why \`application(_:open:)\` and not the legacy \`application(_:openFiles:)\`** — On macOS 13+, AppKit prefers the URL-based modern callback. On macOS 14/15/Tahoe (26.x), the legacy \`openFiles:\` selector is silently skipped when the OS sees that the app has not opted into the modern API. Verified empirically on macOS 26.3.1: \`openFiles:\` was never invoked, \`application(_:open:)\` is.

**Why this doesn't break the existing \`snapzy://\` URL scheme** — \`AppDelegate.applicationWillFinishLaunching\` registers an Apple Event handler for \`kAEGetURL\`. Once that handler is set, AppKit routes URL-scheme events through it directly and does not also call \`application(_:open:)\` for them. The new method's \`isFileURL\` filter is a defensive double-guarantee: even if a future macOS version did surface scheme URLs here, deep links would still be no-ops on this path and continue flowing through the existing handler. Tested: \`snapzy://capture/area\` etc. continue to work unchanged.

**Sandboxing** — Snapzy is sandboxed with \`com.apple.security.files.user-selected.read-write\`. When a user explicitly picks a file via \"Open With\", macOS grants the app implicit, scoped access to that URL, which \`AnnotateWindowController(url:)\` already handles via \`SandboxFileAccessManager.beginAccessingURL\`. No entitlement changes needed.

## Test plan

Verified end-to-end on macOS 26.3.1 (arm64) with a local Debug build:

- [x] \`xcodebuild -scheme Snapzy -configuration Debug build\` succeeds
- [x] After build, \`lsregister -dump\` shows the app's claimed UTIs:
  \`public.png, public.jpeg, public.heic, public.heif, public.tiff, com.compuserve.gif, org.webmproject.webp, public.bmp\`
- [x] \`open -a /Applications/Snapzy.app /path/to/image.png\` opens the annotation editor with that image; diagnostic log confirms:
  \`\`\`
  [INF][ACTION][SnapzyApp.swift:application(_:open:)] Received open-file request {count=1}
  [INF][ACTION][AnnotateManager.swift:openAnnotation(url:)] Annotate window opened for URL image.png
  \`\`\`
- [x] Finder right-click on a PNG → **Open With → Snapzy** → annotation editor opens with that image
- [x] Multi-select 3 images → Open With → Snapzy → 3 annotation windows open
- [x] Existing \`snapzy://capture/area\`, \`snapzy://open/annotate\`, \`snapzy://settings\` deep links still trigger their handlers unchanged
- [x] \`LSHandlerRank = Alternate\` confirmed: Preview remains the default for double-click; Snapzy only shows up in the Open With submenu unless user opts in via \"Always Open With\"

## Notes for reviewer

- The change is intentionally minimal and reuses existing infrastructure — no new managers, no new types.
- Happy to add Vietnamese / Chinese localized strings for the document type name (\"Image\") if the project prefers that; the current \`CFBundleTypeName\` is the internal type label and is not user-visible in macOS 13+ Open With UI, but I can add a localized variant if you'd like consistency with the rest of \`Resources/Localization/\`.
- Future follow-ups (out of scope here, can be separate PRs if useful):
  - Share Extension for Safari / Mail / Photos
  - Services menu entry (\"Edit Image in Snapzy\")

Thanks for building Snapzy! 🙌

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Snapzy/App/SnapzyApp.swift
+++ b/Snapzy/App/SnapzyApp.swift
@@ -38,6 +38,8 @@ struct SnapzyApp: App {
 class AppDelegate: NSObject, NSApplicationDelegate {
   private var coordinator: AppCoordinator?
   private var pendingDeepLinkURLs: [URL] = []
+  private var pendingOpenFileURLs: [URL] = []
+  private var didFinishLaunching = false
 
   func applicationWillFinishLaunching(_ notification: Notification) {
     NSAppleEventManager.shared().setEventHandler(
@@ -57,7 +59,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     let coordinator = AppCoordinator(environment: AppEnvironment.live())
     self.coordinator = coordinator
     coordinator.applicationDidFinishLaunching()
+    didFinishLaunching = true
     flushPendingDeepLinks()
+    flushPendingOpenFileURLs()
   }
 
   func applicationWillTerminate(_ notification: Notification) {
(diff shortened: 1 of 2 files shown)


duongductrong:
Thanks for your contribution, that's great!

PR #226, opened 2026-06-01
feat: add 90° image rotation buttons to annotate editor

omarshahine:
## Summary

Adds rotate-left / rotate-right toolbar buttons next to crop in the annotate editor. Rotates the source image 90° and brings all editor state along with it.

Before this PR there was no way to rotate a screenshot once captured — only crop, draw, blur, mockup, etc.

## What rotates

- Source image, cutout image, and every embedded image layer — via a new `NSImage.rotated90(clockwise:)` helper that preserves Retina pixel resolution and swaps logical point size.
- All annotation bounds + embedded coordinate data (`path`, `highlight`, `line`, `arrow` including elbow/curve control points).
- `cropRect`, `originalCropRect`, and `cutoutAutoAppliedCropRect`.
- `isCropPortraitOrientation` toggles when an aspect-ratio preset (16:9, etc.) is active, so the preset identity survives rotation.
- Embedded image `embeddedImageSourceData` / `embeddedImageSnapshotCacheData` are cleared per asset on rotation so persistence re-encodes from the rotated `NSImage` (otherwise saved files would stale-revert the embedded layers).

## Undo / redo

Introduces a typed undo entry — `enum UndoEntry { case annotations(AnnotationSnapshot); case rotation(RotationSnapshot) }`. The existing annotation-only undo path is unchanged. Rotation gets its own `RotationSnapshot` capturing the 13 mutated fields, so rotation undo can never silently roll back background cutout or crop state (which sit outside the annotation undo stack).

`applyRotationSnapshot` also clears any active crop interaction when restoring — rotation is gated on `!isCropActive`, so a rotation snapshot by definition represents a non-crop state, and a stale `isCropActive` against the restored image dimensions would otherwise corrupt the crop UI.

## Disabled when

- No image is loaded
- Crop is being actively edited (`isCropActive`)
- A background-cutout request is in flight (`isCutoutProcessing`)

## Files

- `Snapzy/Features/Annotate/Services/AnnotateImageRotation.swift` — new helper (NSImage rotation + pure point/rect rotation formulas)
- `Snapzy/Features/Annotate/AnnotateState.swift` — `UndoEntry` enum, `RotationSnapshot`, `rotateImage(clockwise:)`, `canRotateImage`, `rotateAnnotation`, `applyRotationSnapshot`
- `Snapzy/Features/Annotate/Components/AnnotateToolbarView.swift` — two new buttons in `captureToolsGroup`
- `Snapzy/Shared/Localization/L10n.swift` + `Snapzy/Resources/Localization/Features/Annotate.xcstrings` — new `rotateLeft` / `rotateRight` strings
- `SnapzyTests/Features/Annotate/AnnotateImageRotationTests.swift` — 14 new unit tests

## Tests

`xcodebuild test -only-testing:SnapzyTests/AnnotateImageRotationTests -only-testing:SnapzyTests/AnnotateCoreTests` — green.

New tests cover:
- Point rotation formulas (CW + CCW, every corner of an asymmetric image)
- 4× CW identity round-trip
- CW↔CCW reversibility
- Rect rotation formulas (CW + CCW)
- Non-standard rect normalisation
- Full-image rect round-trip
- `NSImage.rotated90` swaps logical point size in both directions
- Two rotations preserve original aspect ratio

## Localisation

English strings added in `L10n.swift` and the Annotate xcstrings catalog. Other locales will fall back to the English default until translated; `swift tools/localization/CatalogTool.swift verify` passes (`keys=1168 missing=0 extra=0`).

## Known limitations

Per-annotation `rotationDegrees` (text/watermark) is clamped to ±45° in `AnnotationProperties.clampedRotationDegrees`, so adding ±90° cleanly isn't possible. The annotation's bounding box rotates correctly; only the inner-text orientation stays relative to the original canvas. Most users rotate before annotating, so this is acceptable for v1.

## Review process

Pre-flight reviewed with `codex review --base master`, which caught the crop-state bug now fixed in `0cfc418`.

### DIFF
--- a/Snapzy/Features/Annotate/AnnotateState.swift
+++ b/Snapzy/Features/Annotate/AnnotateState.swift
@@ -18,6 +18,29 @@ final class AnnotateState: ObservableObject {
     var embeddedImageAssets: [UUID: NSImage]
   }
 
+  /// Snapshot of every piece of state mutated by an image rotation. Used as a dedicated
+  /// undo entry so rotation undo never disturbs the annotation-only undo path.
+  private struct RotationSnapshot {
+    var sourceImage: NSImage?
+    var cutoutImage: NSImage?
+    var isCutoutApplied: Bool
+    var embeddedImageAssets: [UUID: NSImage]
+    var embeddedImageSourceData: [UUID: Data]
+    var embeddedImageSnapshotCacheData: [UUID: Data]
+    var annotations: [AnnotationItem]
+    var cropRect: CGRect?
+    var originalCropRect: CGRect?
+    var cropAspectRatio: CropAspectRatio
+    var isCropPortraitOrientation: Bool
+    var didCutoutAutoApplyCrop: Bool
+    var cutoutAutoAppliedCropRect: CGRect?
+  }
+
+  private enum UndoEntry {
+    case annotations(AnnotationSnapshot)
+    case rotation(RotationSnapshot)
+  }
+
   private struct TextEditingUndoTransaction {
     let annotationId: UUID
     let snapshotBeforeEdit: AnnotationSnapshot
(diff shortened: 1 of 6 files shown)


duongductrong:
Hey @omarshahine, that's great! I have 2 pieces of feedback for this, just want to improve it better for the user (please correct me if I am wrong)

1. The text feels incorrect when rotating 

Snapzy_Recording_2026-06-01_22-02-11

---

2. Please consider moving this rotate annotation item after the sidebar toggle, separated from the sidebar toggle, and crop by adding a divider. What are your thoughts on this?

Snapzy_clipboard_1FF7607B-DC1A-443C-87CD-585247569419

PR #244, opened 2026-06-11
fix: commit first area-selection drag and apply cursor on capture start (#242)

### LINKED ISSUE #242: [Bug]: Command Shift 4 weirdness
### Description

Command Shift 4 leads to an initial crosshair, then a second crosshair remains and I have to repeat my selection drag.

macOS Tahoe 26.5.1

### Steps to Reproduce

1. Command Shift 4
2. crosshair appears
3. drag crosshair over desired area
4. crosshair does not disappear
5. drag crosshair over desired area again
6. finally the screenshot is created and the screenshot preview thumbnail appears

### Expected Behavior

1. Command Shift 4
2. crosshair appears
3. drag crosshair over desired area
4. the screenshot is created and the screenshot preview thumbnail appears

### Actual Behavior

Command Shift 4 leads to an initial crosshair, then a second crosshair remains and I have to repeat my selection drag.

### Screenshots / Recordings

_No response_

### Snapzy Version

1.21.0 (117)

### macOS Version

### Installation Method

DMG download

### Additional Context

_No response_

omarshahine:
## Summary

Fixes two issues when starting an area capture via a global shortcut (e.g. ⌘⇧4) while another app is frontmost. The selection overlay is a `.nonactivatingPanel`, so Snapzy stays in the background during the gesture.

### 1. First drag dropped — fixes #242

The manual-region drag/up were tracked **only** by a *local* `NSEvent` monitor, which fires solely while the app is active. On the first gesture the drag/`mouseUp` land before Snapzy activates, so the local monitor never sees them, the selection never commits, and the overlay resets — exactly the report in #242: *"a second crosshair remains and I have to repeat my selection drag."* The previous "first click" fix (#228) only covered the multi-display lazy-snapshot path, which is unreachable on a single display.

**Fix:** add an observe-only **global** `NSEvent` monitor alongside the local one for `.leftMouseDragged`/`.leftMouseUp`. Global monitors receive events even while the app is inactive, so the first gesture commits. Both monitors are torn down together, and the handlers are idempotent (`updateManualSelection` just records a point; `endManualSelection` early-returns once torn down), so it's safe when both fire while active.

### 2. System arrow lingered over the crosshair until first move

The overlay uses a transparent cursor so only the drawn crosshair shows, but on a background app macOS doesn't apply it until a `cursorUpdate` fires (i.e. the pointer moves).

**Fix:** `NSApp.activate(ignoringOtherApps:)` on session start so the overlay's cursor rects are evaluated immediately. The backdrop is a frozen snapshot, so there is no live window behind the overlay to blur.

## Testing

- Built Debug against the macOS 26.5 SDK — `** BUILD SUCCEEDED **`.
- Verified on a single-display Mac: from another frontmost app, ⌘⇧4 → drag immediately now commits the first gesture, and the crosshair shows with no lingering arrow before moving the mouse.
- Happy path (Snapzy already active) and Esc-cancel still behave.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Snapzy/Services/Capture/AreaSelectionWindow.swift
+++ b/Snapzy/Services/Capture/AreaSelectionWindow.swift
@@ -350,9 +350,16 @@ final class AreaSelectionController: NSObject {
     // The overlay is a non-activating panel, so when the session is triggered from a global
     // shortcut while another app is frontmost, macOS keeps showing the previous app's arrow cursor
     // over our crosshair until the pointer moves and a `cursorUpdate` fires. Activating evaluates
-    // the overlay's cursor rects right away. The backdrop is a frozen snapshot, so there is no live
-    // window behind the overlay to blur.
-    NSApp.activate(ignoringOtherApps: true)
+    // the overlay's cursor rects right away.
+    //
+    // Limit this to frozen-backdrop sessions (screenshot area capture), where a static snapshot
+    // sits behind the overlay so nothing live is dimmed. Backdrop-less sessions — recording-area
+    // selection and the legacy screenshot API — overlay the actual windows being captured, so
+    // activating would deactivate/dim them and leave Snapzy frontmost afterward; those keep the
+    // non-activating behavior this window class is built around.
+    if !selectionBackdrops.isEmpty {
+      NSApp.activate(ignoringOtherApps: true)
+    }
 
     startWindowSelectionPreparationIfNeeded()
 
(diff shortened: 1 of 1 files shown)


duongductrong:
Hi @omarshahine, that's really great! I tested it and noticed one issue: when the user completes a capture, the previously active window requires an extra mouse click to regain focus. I would expect the capture process to finish while keeping the previously active window focused automatically.

Could you help improve this behavior? If so, I think it would be ready to merge and release.

PR #245, opened 2026-06-12
fix: Support both trackpad swipe directions for Quick Access dismiss with natural scrolling

williamcachamwri:
## Summary

Fixes two issues with the two-finger trackpad swipe-to-dismiss gesture in Quick Access.

### 1. Natural scrolling normalization
When macOS "Natural scrolling" is enabled, `scrollingDeltaX` has the opposite sign, causing the swipe direction check to reject valid dismiss gestures. Added `isDirectionInverted` parameter to `QuickAccessTrackpadSwipePolicy.horizontalDelta()` using `event.isDirectionInvertedFromDevice` to normalize the delta.

### 2. Both directions trigger dismiss
Unlike mouse drag (which differentiates swipe-to-dismiss vs. drag-to-app by direction), the two-finger trackpad swipe is used exclusively for dismiss. `dismissTranslation()` now normalizes the accumulated delta via `abs(deltaX) * dismissDirection`, so swiping in either horizontal direction dismisses the card with the visual offset always moving toward the nearest screen edge.

## Changes
- `QuickAccessDraggableView.swift`: Updated `horizontalDelta()` to accept and use `isDirectionInverted` flag
- `QuickAccessDraggableView.swift`: Updated `dismissTranslation()` to allow both swipe directions


duongductrong:
Hey @williamcachamwri,

Thanks for your PR, that's great. I have been tested and figured out a UX issue. That's basically feel un-natural behavior when I'm using two fingers to move left, then the quick access card moving right side, I think it's not a natural feeling, imo. If able, it must be follow the direction of the user action

And is it duplicates when 2 behaviors for one action?

Snapzy_clipboard_7FE3FC06-769F-404B-8F53-D695874274EC

---
So far I think we can improve more because if some users want to move left for X behavior, right for Y behavior, then I think it can definitely be set in Settings for move-left/right/top/bottom for selected behavior, that's depending on every user's habit

duongductrong:
Hey @williamcachamwri,

Thanks for the quick update. I've tested it, and I'm not entirely sure about the current behavior. I noticed that swiping left causes the Quick Access card to move to the right, while swiping right causes it to move to the left. This feels a bit unnatural to me.

As a user, I would expect the card to move in the same direction as my swipe gesture. For example, when I swipe right, the card should move right, and when I swipe left, the card should move left.
 
Snapzy_Recording_2026-06-13_15-36-14

williamcachamwri:
I think users should be given the option to choose between natural swipe and inverted swipe.

duongductrong:
> I think users should be given the option to choose between natural swipe and inverted swipe.

Yes, that's great and flexible. Does it exist in your changes currently?

williamcachamwri:
> > I think users should be given the option to choose between natural swipe and inverted swipe.
> 
> Yes, that's great and flexible. Does it exist in your changes currently?

Not yet, I'm still testing it locally.

duongductrong:
> > > I think users should be given the option to choose between natural swipe and inverted swipe.
> > 
> > 
> > Yes, that's great and flexible. Does it exist in your changes currently?
> 
> Not yet, I'm still testing it locally.

Sounds good. Once it's ready, I'd be happy to test it and compare the user experience between natural and inverted swipe modes. Having both options should cover different user preferences.

williamcachamwri:
Hey @duongductrong,

I added a Trackpad Swipe Direction option in Settings under Quick Access. It has two modes: Natural makes the card follow your finger direction, and Inverted makes the card move the opposite way like scrolling content. Both left and right two-finger swipes now dismiss the card. I tested the build locally and the test suite passed, but it would be great if you could try both modes on your trackpad and confirm the swipe direction feels right before we merge.

duongductrong:
> Hey @duongductrong,
> 
> I added a Trackpad Swipe Direction option in Settings under Quick Access. It has two modes: Natural makes the card follow your finger direction, and Inverted makes the card move the opposite way like scrolling content. Both left and right two-finger swipes now dismiss the card. I tested the build locally and the test suite passed, but it would be great if you could try both modes on your trackpad and confirm the swipe direction feels right before we merge.

That's great now for the UX behavior, just fix a small issue here

Snapzy_Recording_2026-06-13_16-09-34

williamcachamwri:
Hmm, that's weird, I didn't encounter that problem on my Mac.

williamcachamwri:
https://github.com/user-attachments/assets/eb43bd2c-2e37-4d4d-8430-86a88b376507

williamcachamwri:
@duongductrong,
I also fixed the small UX issue where the hover overlay and corner buttons would appear while swiping the card. They are now hidden during an active swipe so they do not visually interfere with the gesture.

williamcachamwri:
Hey @duongductrong, 
Have you tested it yet? I just tested it and the OCR Text Capture initially failed to recognize the text, but later it worked normally.

williamcachamwri:
Hmmm, maybe it's working normally, it's probably fine...

duongductrong:
Hey @williamcachamwri, Is this PR ready for merge, any changes bro?

williamcachamwri:
It was fine and nothing changed.

PR #256, opened 2026-06-19
Fix Quick Access dismiss lag by deferring cleanup

### LINKED ISSUE #219: [Bug]: Quick Access edit/close buttons have ~1s lag (main-thread blocking I/O)
### Description

Two Quick Access actions feel sluggish: clicking the pencil (edit) on a card takes ~1s+ before the Annotate window appears (worst on the first click of the session), and clicking close / swipe-dismiss leaves the card frozen for ~1s before the dismiss animation runs. The rest of the app stays responsive — the lag is isolated to these two handlers. Both root-cause to synchronous I/O running on the main actor before SwiftUI can render the response. Detailed root cause + suggested fix direction in the Additional Context section below.

### Steps to Reproduce

1. Take any screenshot so a Quick Access card appears in the corner.
2. Click the pencil icon on the card — note the delay before the Annotate window appears (first click of the session is the worst).
3. Take another screenshot.
4. Click the close button on that card (or swipe it toward the screen edge) — note the delay before the card animates away.

### Expected Behavior

Both interactions should feel instant. The Annotate window should appear within ~50 ms of clicking the pencil (a placeholder is fine while the full image finishes loading). The card should start its dismiss animation within one frame (~16 ms) of clicking close.

### Actual Behavior

The Annotate window takes ~1 s+ to appear after clicking the pencil. The card visibly freezes for ~1 s after clicking close before the dismiss animation kicks in.

### Screenshots / Recordings

_No response_

### Snapzy Version

1.20.0

### macOS Version

macOS 15 (Sequoia)

### Installation Method

Built from source

### Root cause

Both handlers do significant work **synchronously on the main actor** before SwiftUI can render the response.

**Edit button** — `AnnotateManager.openAnnotation(for:)` (`Snapzy/Features/Annotate/AnnotateManager.swift`) runs this chain on the main thread before the window is shown:

- `NSApp.setActivationPolicy(.regular)` in `becomeRegularApp()` — Launch Services / Dock / Cmd+Tab registration on the first call per session (~100–500 ms). Subsequent calls are no-ops thanks to `isRegularAppMode`, which matches the "first edit is slowest" feel.
- `AnnotationSessionStore.shared.load(for:)` — when a sidecar exists, synchronously reads `manifest.json` + the full original PNG (`original.bin`) + optional cutout + embedded assets.
- `AnnotateWindowController.init` then calls `NSImage(contentsOf:)` (full decode) **and** `Data(contentsOf:)` (raw bytes for the session cache) on the main thread.
- `NSHostingView(rootView: AnnotateMainView(state:))` builds the full Annotate SwiftUI tree (toolbar, properties bar, canvas, sidebar) up-front.

Cumulative first-click cost: roughly 300 ms – 1.5 s.

**Close button** — `QuickAccessManager.removeItem(id:)` (`Snapzy/Features/QuickAccess/QuickAccessManager.swift`) runs all of this before the `withAnimation { items.removeAll }` at the end:

- `CaptureHistoryStore.shared.hasRecord(forFilePath:)` — synchronous GRDB `dbPool.read { fetchCount(db) }`.
- `AnnotationSessionStore.shared.deleteSession(for:)` — synchronous `removeItem(at:)` on the sidecar package directory.
- `pinWindowManager.close(id:)` — synchronous `NSWindow.close()` (cheap when unpinned).

The dismiss animation only kicks in after all of that returns. Even `isDismissing = true` (set first in `dismissItem`) doesn't paint until the chain unwinds, because SwiftUI batches state changes per call stack.

For reference: `DiagnosticLogger.log` and `QuickAccessSound.play` are **not** contributors — both already dispatch to background queues.

### Suggested fix direction

Same pattern for both: **let the UI respond first, do the I/O after.**

- **Edit**: split `AnnotateWindowController.init` into a sync phase (build state with the QA card's in-memory thumbnail as a placeholder, show the window) and an async phase (`Task.detached(priority: .userInitiated)` for sidecar load + image decode + raw bytes, then hop back to the main actor to populate `AnnotateState.sourceImage`). Leave `becomeRegularApp` where it is — its first-call cost is one-time per session.
- **Close**: reorder `removeItem` so the timer cancel + `withAnimation` happen synchronously, then move the SQLite check + sidecar delete + temp file delete into a `Task.detached` that captures the URL up front.

`Task.detached(priority: .userInitiated)` is already used for save-and-close in `AnnotateWindowController` — the same shape fits here.

### Things to keep an eye on

- The "preserve temp file if a history record exists" logic in `removeItem` must stay correct after the move to async — it's covered by `SnapzyTests/Features/QuickAccess/QuickAccessHistoryCleanupTests.swift`, and those tests will need to `await` the deferred I/O.
- Fast open→close on the editor: the in-flight async load needs to be cancellable from `windowShouldClose` / `deinit` (store a `Task` handle on the controller).
- The placeholder (in-memory thumbnail, ~200 px) will look blurry for ~100–300 ms during async load — a small loading affordance is probably worth adding.
- Worth a Time Profiler trace before & after to confirm the fix lands on the actual bottleneck.

### Environment notes

- macOS 26.4.1 (Tahoe) — not in the dropdown.
- Reproduced against current `master` (`MARKETING_VERSION = 1.20.0`); the relevant code paths are unchanged in the latest release `v1.20.1`.
- I won't be able to take this on myself (uni exams). Leaving unassigned — happy for someone in the community to pick it up.

jjoanna2-debug:
## Summary
- Remove Quick Access cards from UI before temp/history/session cleanup runs.
- Defer dismiss cleanup until after SwiftUI can commit the removal animation.
- Yield before opening annotation from the edit action so the button response is not blocked in the same call stack.
- Add coverage that temp-file cleanup happens after synchronous card removal.

Fixes #219.

## Verification
- `xcodebuild -project Snapzy.xcodeproj -scheme Snapzy -configuration Debug -destination 'platform=macOS' -derivedDataPath build/DerivedData -clonedSourcePackagesDirPath build/SourcePackages CODE_SIGN_IDENTITY= CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO build`\n- `git diff --check`\n\nNote: I also attempted `./scripts/run-tests.sh -only-testing:SnapzyTests/QuickAccessHistoryCleanupTests`; it built and linked, then the app-hosted XCTest run hung after launching `Snapzy Debug.app`. No leftover app/test/build processes remained after cleanup.

### DIFF
--- a/Snapzy/Features/QuickAccess/Components/QuickAccessCardView.swift
+++ b/Snapzy/Features/QuickAccess/Components/QuickAccessCardView.swift
@@ -383,7 +383,10 @@ struct QuickAccessCardView: View {
   }
 
   private func openAnnotation() {
-    AnnotateManager.shared.openAnnotation(for: item)
+    Task { @MainActor in
+      await Task.yield()
+      AnnotateManager.shared.openAnnotation(for: item)
+    }
   }
 
   private func openVideoEditor() {
--- a/Snapzy/Features/QuickAccess/QuickAccessManager.swift
+++ b/Snapzy/Features/QuickAccess/QuickAccessManager.swift
@@ -416,11 +416,50 @@ final class QuickAccessManager: ObservableObject {
       return
     }
 
-    // Auto-delete temp files on dismiss (unsaved captures)
-    // Skip deletion if history is enabled and the file has a history record —
-    // the retention service will clean it up when the record ages out.
-    if tempCaptureManager.isTempFile(item.url) {
-      let url = item.url
+    let url = item.url
+    let isTempFile = tempCaptureManager.isTempFile(url)
+
+    cancelDismissTimer(for: id)
+    pinWindowManager.close(id: id)
+    editingItemIds.remove(id)
+    activityHoldItemIds.remove(id)
+    // Clear annotation session cache for this item
+    AnnotateManager.shared.clearSessionData(for: id)
+    // Fast animation (0.15s) for immediate perceived response
+    withAnimation(.spring(response: 0.15, dampingFraction: 0.8)) {
+      items.removeAll { $0.id == id }
+    }
+
+    if items.isEmpty {
+      panelController.hide()
+    }
+
+    scheduleDismissCleanup(for: url, isTempFile: isTempFile)
+  }
+
+  /// Remove a screenshot from the stack (backward compatible alias)
+  func removeScreenshot(id: UUID) {
+    removeItem(id: id)
+  }
+
+  private func scheduleDismissCleanup(for url: URL, isTempFile: Bool) {
+    guard isTempFile else {
+      DiagnosticLogger.shared.log(
+        .info,
+        .action,
+        "Quick access item dismissed; saved file retained",
+        context: ["fileName": url.lastPathComponent]
+      )
+      return
+    }
+
+    // Let SwiftUI commit the card removal before checking history or touching disk.
+    Task { @MainActor in
+      await Task.yield()
+
+      // Auto-delete temp files on dismiss (unsaved captures).
+      // Skip deletion if history is enabled and the file has a history record —
+      // the retention service will clean it up when the record ages out.
       let historyEnabled = UserDefaults.standard.bool(forKey: PreferencesKeys.historyEnabled)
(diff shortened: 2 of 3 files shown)


duongductrong:
Thanks @jjoanna2-debug, LGTM!

PR #278, opened 2026-06-24
fix: normalize export_location to ~/ relative path on export

st1020:
## Problem

The exported `config.toml` always wrote `export_location` as an absolute path (e.g. `/Users/stone/Desktop`). This makes the config file machine-specific and harder to read or share across accounts.

## Solution

Collapse the home directory prefix to `~/` when writing `export_location` during export, so the field always reads like `~/Desktop` regardless of the actual username.

### Changes

- **`SnapzyConfigurationPaths`** — add `collapsingHomePath(_:)` and `collapsingHomePath(_:homeDirectory:)` as the inverse of the existing `expandedUserPath` helpers.
- **`SnapzyConfigurationExporter`** — apply `collapsingHomePath` when writing the `export_location` field.
- **`SnapzyConfigurationPathsTests`** — add `testCollapsingHomePathConvertsAbsolutePathToTilde` covering subdirectory, exact home root, and out-of-home cases.

## Compatibility

The import side already expands `~/` via `expandedUserPath`, so the roundtrip (export → import → export) is stable. Users with existing config files that contain absolute paths will have the field silently upgraded to the `~/` format on the next automatic sync (same behaviour as any other settings change that results in a clean overwrite).

### DIFF
--- a/Snapzy/Services/Configuration/SnapzyConfigurationExporter.swift
+++ b/Snapzy/Services/Configuration/SnapzyConfigurationExporter.swift
@@ -34,7 +34,7 @@ enum SnapzyConfigurationExporter {
     writer.value("appearance", appearance(defaults: defaults))
     writer.value("play_sounds", defaults.object(forKey: PreferencesKeys.playSounds) as? Bool ?? true)
     writer.value("start_at_login", LoginItemManager.isEnabled)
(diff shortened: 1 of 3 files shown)


duongductrong (approved):
LGTM

PR #285, opened 2026-06-25
Fix OCR capture region offset

didogakki:
Summary
- Fix the OCR capture region coordinate conversion so selections are offset correctly in screen space.
- Keep the captured OCR region aligned with the user-selected rectangle.

Validation
- git diff --check
- xcodebuild -project Snapzy.xcodeproj -scheme Snapzy -configuration Debug -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO build


duongductrong:
This PR doesn't have any updates, so we decided to close it. Thanks for your effort on this contribution

PR #294, opened 2026-06-26
Add space-to-move for area capture selection

vxirau:
Hey! 👋

I use Snapzy all day for screenshots, and the one thing I kept reaching for out of habit was being able to hold space to reposition the selection while I'm dragging it out, the way the built-in macOS screenshot tool (⌘⇧4) works. Snapzy didn't have it, so I added it.

While you're dragging out a capture region, hold Space and the whole selection box moves with your cursor (same size, just repositioned). Let go of space and you're back to resizing from wherever it ended up. It's that small thing where you start the drag, realize the corner's slightly off, and now you can nudge it into place instead of starting over.

It's all in AreaSelectionController (in AreaSelectionWindow.swift), which already tracks the live drag:
- While space is held, both corners of the selection get shifted by the pointer delta instead of just the dragged corner, so the size stays put and the box follows the mouse.
- I added local + global key monitors for the space bar alongside the drag monitors that were already there, so it works even when the overlay is triggered from a global shortcut (and the local one swallows the space key so there's no system beep).
- It only kicks in during normal region selection, the window-picker mode and the Escape/right-click-to-cancel behavior are untouched.

Totally open to feedback if you'd want it done differently, happy to adjust. Thanks for the great app! 🙏

### DIFF
--- a/Snapzy/Services/Capture/AreaSelectionWindow.swift
+++ b/Snapzy/Services/Capture/AreaSelectionWindow.swift
@@ -88,6 +88,10 @@ final class AreaSelectionController: NSObject {
   /// app activates, so the local monitor never sees them and the selection silently resets.
   /// A global monitor still receives those events, ensuring the first gesture commits.
   private var manualSelectionGlobalMonitor: Any?
+  private var manualSelectionKeyLocalMonitor: Any?
+  private var manualSelectionKeyGlobalMonitor: Any?
+  private var isMovingManualSelection = false
+  private var manualSelectionLastPointerLocation: CGPoint?
   private var previouslyActiveApplication: NSRunningApplication?
 
   /// Whether the overlay should be dismissed immediately after a selection is made.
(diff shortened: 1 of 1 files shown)


duongductrong:
Hey @vxirau,

Thanks for your PR, that's great, but I wanna know why you're choosing "space" to design the "move" behavior? Just because the space in the default behavior is switching between capture selected window or portion. And if you wanna adjusting the select area before capture, why don't we use "Capture Markup" (cmd+shift+7)

Please share your thoughts. Thank you so much.

vxirau:
Hey @duongductrong , thanks for taking a look.

The reason I chose Space is mainly because this targets a slightly different moment in the capture flow than Capture Markup.

Capture Markup is useful after the screenshot has already been taken, when you want to edit or move annotations/content in the captured image. This PR is about adjusting the capture region before the screenshot is taken. Sometimes you start dragging the area and realize the whole box is just a little offset. Without this, you have to cancel or release and start the selection again. Holding Space lets you keep the same size and just reposition the box before committing the capture.

I chose Space because it matches the native macOS screenshot behavior with `⌘⇧4`: while dragging a region, holding Space moves the current selection instead of resizing it. So it should feel familiar to users who already use the built-in screenshot tool.

It also only applies while actively dragging out a region selection, so it should not affect the window-picker flow or the existing capture mode behavior. Happy to adjust the implementation if you prefer a different key or want stricter scoping.

duongductrong (approved):
lgtm

PR #295, opened 2026-06-26
Add a keyboard shortcut to edit the latest capture from its preview

vxirau:
Hey again 👋 

This one's about keeping the capture flow fully keyboard-driven. Right now, after you take a screenshot the Quick Access preview pops up in the corner, but to edit it you have to grab the mouse and click the edit button before it disappears. I wanted to never leave the keyboard, so I added a shortcut that opens the editor for the most recent capture while its preview is still on screen.

What it does
- Take a capture → while its Quick Access preview is showing → press the shortcut → the editor opens on that capture. No mouse needed.
- Works for both screenshots (annotation editor) and recordings (video editor).
- Configurable in Settings → Shortcuts → Quick Access ("Edit latest capture"), default ⌘↩.

Design notes
- It's a contextual shortcut, not a global hotkey, the key monitor is only installed while a preview is actually on screen, so the combo is never stolen from other apps the rest of the time.
- A global monitor is used (alongside a local one) because the preview is a non-activating panel that can't take key focus, right after a capture another app still holds focus. The local monitor consumes the matched event so there's no system beep when Snapzy itself is frontmost.
- Storage/persistence/matching follows the existing AnnotateShortcutManager action-shortcut pattern, and the settings row reuses the shared ShortcutRecorderView, so it fits right in with the other shortcuts (recordable, clearable, resettable, and included in "Reset to defaults").
- New strings go through the existing L10n helper with English defaults, so untranslated locales fall back gracefully.

Happy to tweak the default shortcut or wording if you'd prefer something else :)

### DIFF
--- a/Snapzy/Features/Annotate/Managers/AnnotateWindow.swift
+++ b/Snapzy/Features/Annotate/Managers/AnnotateWindow.swift
@@ -33,8 +33,6 @@ extension Notification.Name {
 /// Custom NSWindow for annotation editing with dark mode appearance
 class AnnotateWindow: NSWindow {
   weak var interactionState: AnnotateState?
-  /// Invoked when Esc is pressed while not editing text or cropping. Returns true if it handled the
-  /// key (e.g. discarded the capture and closed the window), in which case the event is consumed.
   var onEscape: (() -> Bool)?
   private static let activeEditorLevel = NSWindow.Level(rawValue: NSWindow.Level.floating.rawValue + 1)
   private var restingLevel: NSWindow.Level = .normal
--- a/Snapzy/Features/Annotate/Managers/AnnotateWindow.swift
+++ b/Snapzy/Features/Annotate/Managers/AnnotateWindow.swift
@@ -33,6 +33,9 @@ extension Notification.Name {
 /// Custom NSWindow for annotation editing with dark mode appearance
 class AnnotateWindow: NSWindow {
   weak var interactionState: AnnotateState?
+  /// Invoked when Esc is pressed while not editing text or cropping. Returns true if it handled the
+  /// key (e.g. discarded the capture and closed the window), in which case the event is consumed.
+  var onEscape: (() -> Bool)?
   private static let activeEditorLevel = NSWindow.Level(rawValue: NSWindow.Level.floating.rawValue + 1)
   private var restingLevel: NSWindow.Level = .normal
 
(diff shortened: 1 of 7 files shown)


duongductrong:
Hey @vxirau,

After manual testing, I have two pieces of feedback for this PR
- Escaping the Annotate Window, then the Quick Access Card does not reappear at its position, while that's my expectation from old behavior
- While trying to add some annotated items and save (cmd+s) on the Annotate Window, I also expect the annotate item reappear at their positions. However, this PR makes it's gone

vxirau:
Hey @duongductrong 👋

Thanks for the thorough testing, both fair points.

1) Esc → Quick Access card reappears: done. Esc now closes the Annotate window without discarding the capture, so the card comes back at its position (with its auto-dismiss timer reset so it doesn't instantly vanish), and you can re-open it again with the shortcut.

2) ⌘S → card behavior: here I made a deliberate choice I'd like to run by you. I treat ⌘S as a "commit & finish" action — it bakes the annotations in, saves to the default screenshot folder, and then dismisses the card, the idea being that once you've deliberately pressed save you're done with that capture, so clearing it from the corner feels natural. Esc is then the "keep it around / not done yet" path.

I realize this differs from the old behavior where the card stayed after saving, so I'm very happy to flip ⌘S back to keep the card if you'd prefer to preserve that, small change. Just let me know which you'd rather have.

duongductrong:
Hey @vxirau,

I understand your point 2, and I actually implemented it that way before. However, after using it for a while, I felt it wasn't very user-friendly.

The main reason is that adding annotation items in the Annotate Window is often just a temporary action. If the user chooses **Drag to App**, the workflow ends there. Otherwise, I think annotations should behave like part of the current Quick Access session.

For example, if **After Capture > Save** is enabled and **Toggle Off** is selected, the user can take a screenshot, add annotations, and then press **Save**. If the card disappears immediately after saving, users may have no idea where their screenshot went. This can be confusing because they often expect to continue interacting with it during the current session. Basically, we don't want to interrupt the user's workflow when they close or save the Annotate window. It's just one step in the workflow centered around the Quick Access panel

If this is a feature that many people really want, I think it would be better as an optional setting rather than changing the default behavior that users have already learned. Otherwise, users with **After Capture > Save > Toggle Off** enabled may suddenly find that their screenshot disappears after pressing **Save**, without knowing where to find it afterward.

vxirau:
You're right, and thanks for laying out the reasoning. I went back and kept ⌘S consistent with the existing "Done" behavior, it saves the annotations and leaves the Quick Access card in place for the rest of the session, so nothing disappears out from under the user.

You mentioned it could live behind an optional setting; I decided not to add one for now and to just keep the default people already know, rather than introduce another toggle for an edge case. Easy to revisit if it turns out people want it.

Esc still closes the editor and brings the card back at its position (with a fresh auto-dismiss countdown), and ⌘ + enter reopens it.

Let me know what you think.

duongductrong (approved):
lgtm, thanks @vxirau

duongductrong:
Hey @vxirau,

Just a quick note: I decided to disable the "Edit Latest Capture" shortcut by default because I think it could accidentally intercept users pressing ⌘ + Enter in the active window. In my opinion, that could become pretty annoying, so it feels safer to leave it off by default and let users enable it if they want.

Snapzy_clipboard_28FD4C84-C1E8-4523-BE8D-E6C0364636DC

PR #296, opened 2026-06-26
Fix Simplified and Traditional Chinese translations for Snapzy’s screen recording feature

j178:
In Chinese, `录音` / `錄音` specifically means **audio recording**, such as recording from a microphone. It does not normally mean recording the screen.

In Snapzy, `Recording` refers to screen/video recording. Translating this as `录音` / `錄音` makes the UI read as if the feature is only for recording sound.

Native Chinese UI usually uses:

- Simplified Chinese: `屏幕录制` or the shorter `录屏`
- Traditional Chinese: `螢幕錄製` or the shorter `錄屏`

This PR updates the affected Chinese strings to use screen-recording terminology, while keeping audio/microphone-related text distinct where needed.

### DIFF
--- a/Snapzy/Resources/Localization/Features/Capture.xcstrings
+++ b/Snapzy/Resources/Localization/Features/Capture.xcstrings
@@ -835,13 +835,13 @@
         "zh-Hans": {
           "stringUnit": {
             "state": "translated",
-            "value": "录音"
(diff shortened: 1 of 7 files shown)


duongductrong (approved):
LGTM

PR #298, opened 2026-06-27
Add optional live (non-frozen) mode for area capture

vxirau:
I noticed an inefficiency when capturing screenshots: the moment you trigger an area capture, Snapzy grabs and renders a full-screen snapshot just to show a frozen backdrop to select against. It works great for precision, but it does a chunk of work up front every single time, and it has a side effect I kept running into, which is that anything moving on screen (a playing video, an animation) freezes the instant I start selecting, so I can't line up the exact frame I actually want.

I understand the freeze is genuinely useful for some use cases, so I didn't change the default, I just added the option to disable it. There's a small toggle in Settings → Capture → Screenshot, "Freeze screen during area capture", on by default so nothing changes for anyone unless they want it to. Turn it off and the screen just stays live while you drag out your selection, and the region gets captured the moment you release.

The nice part is that the live path is actually lighter than the default, it skips that up-front snapshot-and-render entirely and reuses the capture path that was already in the codebase as a fallback, and the selection overlay is already excluded from the capture, so nothing leaks into the image. The only real tradeoff is that the frame comes from when you release rather than when you trigger, which is exactly what you want when the whole point is to catch a live moment.

### DIFF
--- a/Snapzy/Features/Capture/CaptureViewModel.swift
+++ b/Snapzy/Features/Capture/CaptureViewModel.swift
@@ -138,6 +138,13 @@ final class ScreenCaptureViewModel: ObservableObject, KeyboardShortcutDelegate {
     UserDefaults.standard.object(forKey: PreferencesKeys.screenshotShowCursor) as? Bool ?? false
   }
 
+  /// When enabled (default), area capture selects against a frozen snapshot of the screen. When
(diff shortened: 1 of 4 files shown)


duongductrong:
Awesome work, Thanks for your effort on this PR @vxirau, I actually plan to add the non-frozen back 😆

PR #299, opened 2026-06-27
Fix Escape cancel on pre-record toolbar

rv4no:
## Summary
- make the pre-record toolbar a non-activating key panel so it can receive Escape without stealing app focus
- handle Escape/cancelOperation directly in pre-record mode and route it through the existing cancel callback

## Verification
- xcodebuild -project Snapzy.xcodeproj -scheme Snapzy -destination 'platform=macOS' -skipPackagePluginValidation CODE_SIGNING_ALLOWED=NO build\n\n## Notes\n- no local build artifacts, debug app bundles, zip files, or packaging-only changes are included

### DIFF
--- a/Snapzy/Features/Recording/RecordingCoordinator.swift
+++ b/Snapzy/Features/Recording/RecordingCoordinator.swift
@@ -248,6 +248,12 @@ final class RecordingCoordinator: ObservableObject {
       "isActive": "\(isActive)",
       "recorderState": "\(recorder.state)",
     ])
+
+    guard recorder.state != .idle else {
+      cleanup()
+      return
+    }
+
     Task {
       await recorder.cancelRecording()
       cleanup()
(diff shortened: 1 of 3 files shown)


duongductrong:
Hey @rv4no,

Thanks for your effort. I've been testing the improvement of this PR. First of all, I can't press interacting pre-record toolbar anymore

rv4no:
@duongductrong Good catch. I had mainly tested the app/window selection path, where this did not reproduce. The issue was in manual region selection: the crop overlay and pre-record toolbar coexist, and the overlay could keep stealing interaction/focus.

Updated the fix so the toolbar reliably receives clicks/Escape, re-fronts after overlay creation, and pre-record cancel cleans up immediately when idle. Added focused regression tests for toolbar clickability/window style and Escape cancel.

rv4no:
Updated:

Current fix is still a bit shaky. So I'll keep it draft for now.

The root issue is that the pre-record toolbar can be visible but not actually key/frontmost after area selection, because the previously active app may regain focus. That makes Esc unreliable and can also cause the first click on the toolbar to miss.

PR #302, opened 2026-06-27
Add option to hide menu bar icon

vxirau:
Hey again!

Small follow-up PR. Since Snapzy can already be driven mostly from shortcuts, I wanted to make the menu bar icon optional for people who prefer a cleaner menu bar, especially considering macOS 27 makes it harder for third-party menu bar managers to operate.

What it does:
- Adds a new option in Settings → General: “Show menu bar icon”
- When disabled, Snapzy removes its menu bar status item
- If the icon is hidden and Snapzy is already running, opening the app again brings up Settings so there’s still an easy way back in
- The default stays unchanged: the menu bar icon is visible unless the user turns it off

This is intentionally a small preference-level change, not a behavior change for existing users. It keeps the current menu bar workflow by default, but lets keyboard-heavy users hide the icon once they have their shortcuts set up.

Happy to adjust the wording or placement if you’d prefer it somewhere else.

Thanks!

### DIFF
--- a/Snapzy/App/AppStatusBarController.swift
+++ b/Snapzy/App/AppStatusBarController.swift
@@ -52,7 +52,7 @@ final class AppStatusBarController: ObservableObject {
     self.updater = updater
     self.didDetectCrash = didCrash
 
-    setupStatusItem()
+    syncStatusItemVisibility()
     buildMenu()
     observeRecordingState()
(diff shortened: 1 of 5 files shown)


duongductrong (approved):
lgtm

PR #306, opened 2026-06-28
fix: support Fn key in shortcut recording and matching

### LINKED ISSUE #305: [Bug]: Fn keys not recognized
### Description

When trying to change the keyboard shortcuts, the fn keys don't seem to be recognized, the "press keys" don't register my "fn-key + fn2" (or other fn key) press

macOS Tahoe 26.5
MBP

### Steps to Reproduce

1. Open settings
2. Shortcuts tab
3. Change shortcut for any capture item
4. Press fn key + F2,F3,F4 etc

### Expected Behavior

At step 4, it should register the F2/Fn

### Actual Behavior

It waits in "press keys" mode

### Screenshots / Recordings

https://github.com/user-attachments/assets/c18d6f6d-c67a-4bd8-b066-49935e11f9ee

### Snapzy Version

Version 1.26.0 (128)

### macOS Version

### Installation Method

None

### Additional Context

_No response_

williamcachamwri:
## What changed

Added Fn (`.function`) modifier support across the shortcut system so users can record and use the Fn/globe key in keyboard shortcuts.

### Changes

**`KeyboardShortcutManager.swift` — `ShortcutConfig`**
- Added `functionCarbonModifier` constant (`0x2000`) — Carbon has no native Fn constant, so we use a custom bit internally
- `init(from:event:)` now recognizes `.function` modifier flag from NSEvent
- `displayString` / `displayParts` render `"fn"` when Fn modifier is set
- `keyCodeToString` handles `0x3F` (kVK_Function / globe key) → `"fn"`
- `registerShortcutIfNeeded` / `registerOverlayShortcutIfNeeded` strip the Fn bit before Carbon registration and skip if Fn was the only modifier (preventing a modifier‑less hotkey that would fire on every key press)

**`CaptureOverlayShortcutSettings.swift`**
- `matches(_:)` — removed `guard !event.modifierFlags.contains(.function)` that rejected Fn‑modified events
- `isAllowedSingleKey` — removed `.function` from the forbidden‑modifiers list so Fn+letter can be recorded as a single‑key shortcut

**`AnnotateShortcutManager.matchesShortcut(_:event:)`**
- Added `.function` flag comparison to properly match annotated action shortcuts that include Fn

### Limitations
- Carbon `RegisterEventHotKey` does not support Fn as a modifier. When Fn is the only modifier (e.g., `Fn+F1`), the shortcut is displayed but **not** registered globally (registering with zero modifiers would fire on every key press). When Fn is combined with other modifiers (e.g., `Fn+Cmd+F1`), the shortcut is registered with the other modifiers only — Fn is stripped.

### Testing
- [x] Build succeeds via `xcodebuild`
- [x] Debug app launches and runs
- [x] Shortcut recorder in Settings now accepts Fn + key presses
- [x] Display shows `fn` keycap correctly

Closes #305

### DIFF
--- a/Snapzy/Features/Annotate/Services/AnnotateShortcutManager.swift
+++ b/Snapzy/Features/Annotate/Services/AnnotateShortcutManager.swift
@@ -347,6 +347,7 @@ final class AnnotateShortcutManager: ObservableObject {
     if config.modifiers & UInt32(shiftKey) != 0 { expected.insert(.shift) }
     if config.modifiers & UInt32(optionKey) != 0 { expected.insert(.option) }
     if config.modifiers & UInt32(controlKey) != 0 { expected.insert(.control) }
+    if config.modifiers & ShortcutConfig.functionCarbonModifier != 0 { expected.insert(.function) }
     return flags == expected
   }
 }
--- a/Snapzy/Services/Capture/CaptureOverlayShortcutSettings.swift
+++ b/Snapzy/Services/Capture/CaptureOverlayShortcutSettings.swift
@@ -55,7 +55,6 @@ struct CaptureOverlayShortcut: Equatable, Codable {
   }
 
   func matches(_ event: NSEvent) -> Bool {
-    guard !event.modifierFlags.contains(.function) else { return false }
     guard UInt32(event.keyCode) == keyCode else { return false }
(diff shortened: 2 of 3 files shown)


duongductrong (approved):
lgtm, thanks @williamcachamwri

PR #321, opened 2026-07-01
Fix "stroke" localization strings for Spanish and French

jmcubel:
Fix "stroke" localization.

Correct the Spanish and French translations to use the drawing meaning of "stroke" instead of the medical one.


duongductrong (approved):
lgtm

PR #336, opened 2026-07-06
[codex] Fix screenshot overlay drag cursor handling

omarshahine:
## Summary
- resolve the branch against current `master`
- keep screenshot overlays non-activating so foreground-window capture still targets the app that was frontmost when capture started
- use cursor rect refresh plus pointer tracking for crosshair ownership instead of activating Snapzy

## Root Cause
The conflicted upstream code fixed first-drag handling with global monitors, but also activated Snapzy for frozen screenshot overlay sessions. That activation can make Snapzy the foreground app during capture setup/teardown, which risks foreground-window capture selecting the wrong app/window.

## Validation
- `xcodebuild -project Snapzy.xcodeproj -scheme Snapzy -configuration Debug -derivedDataPath .build/xcode-derived-data CODE_SIGN_IDENTITY= CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO build`
- `xcodebuild -project Snapzy.xcodeproj -scheme Snapzy -configuration Debug -destination platform=macOS -derivedDataPath .build/xcode-test-derived-data CODE_SIGN_IDENTITY= CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO -only-testing:SnapzyTests/AreaSelectionModelsTests test`

## Notes
- `./scripts/run-tests.sh -only-testing:SnapzyTests/AreaSelectionModelsTests` fails locally before compile because this machine does not have the configured Mac Development signing certificate.

### DIFF
--- a/Snapzy/Services/Capture/AreaSelectionWindow.swift
+++ b/Snapzy/Services/Capture/AreaSelectionWindow.swift
@@ -451,13 +451,16 @@ final class AreaSelectionController: NSObject {
     // Activate pooled windows (instant show)
     activatePooledWindows()
 
-    // Keep the overlay non-activating so foreground-window capture still observes the app that was
-    // frontmost when the capture started. Cursor rect refresh plus pointer tracking gives the
-    // crosshair ownership without making Snapzy the active app.
+    // Keep live overlay sessions non-activating so foreground-window capture still observes the app
+    // that was frontmost when the capture started. Cursor rect refresh plus pointer tracking gives
+    // backdrop-less sessions crosshair ownership without making Snapzy the active app. Frozen
+    // sessions already render over a captured backdrop and do not need this key-window churn.
     for (_, window) in windowPool {
       window.invalidateCursorRects(for: window.overlayView)
     }
-    startPointerTrackingIfNeeded()
+    if selectionBackdrops.isEmpty {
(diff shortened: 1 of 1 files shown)


omarshahine:
@duongductrong Would you mind taking a look at this PR when you get a chance? It fixes screenshot overlay drag cursor handling and frozen area selection pointer tracking. All checks are passing. Thanks!

duongductrong (approved):
Thanks for the fix, @omarshahine — and sorry for the slow review.

The reasoning checks out on my side: `NSApp.activate(ignoringOtherApps:)` during frozen overlay sessions makes Snapzy frontmost, and both `CaptureOutputNaming.fromFrontmostApp()` and `ActiveWindowResolver` read `NSWorkspace.frontmostApplication` — so the activation could genuinely leak into capture context/window resolution. Replacing it with cursor-rect invalidation + pointer tracking keeps the overlay non-activating, which matches what `AreaSelectionWindow` is designed around, and the global monitors already cover the first-drag commit case. Removing the now-dead `previouslyActiveApplication` restore is consistent with that.

Diff is minimal, merges cleanly against current `master`, and CI (build + swiftformat) is green. Approved — will merge.

duongductrong (approved):
lgtm

PR #347, opened 2026-07-08
feat: Detect hyperlinks in OCR captured text and offer to open them

omarshahine:
## Summary

After **Capture Text (OCR)**, any web links found in the recognized text (including QR code URL payloads) now surface a CleanShot-style floating prompt with clickable rows that open the link in the default browser.

- **`OCRLinkDetector`** — `NSDataDetector`-based extraction on the composed clipboard text. Only http/https links are surfaced (no `mailto:` or custom schemes, so nothing unexpected can launch); bare domains like `example.com` count. Dedupes case and trailing-slash variants — scheme/host only, paths stay case-sensitive — and caps the prompt at 3 links to keep it compact.
- **`OCRLinkPromptManager`** — interactive non-activating `NSPanel` styled to match the existing toast system, positioned above the bottom-center toast slot so it never overlaps the "Copied to Clipboard" toast. Auto-dismisses after 10 s, paused while hovering. Detection is passive: nothing opens without an explicit click.
- **Preferences** — new "Detect Links" toggle in Capture → OCR section (`ocr.linkDetectionEnabled`, default on).
- **Localization** — all new strings translated in the 10 catalog languages (insert-only diff in `Capture.xcstrings`).

## Testing

- 9 unit tests in `OCRLinkDetectorTests` (detection, bare-domain promotion, scheme filtering, dedupe semantics, ordering/limit, display formatting) — all pass.
- Full `Snapzy` app target builds against current `master`; existing `MediaCoreTests` suite still passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Snapzy/Resources/Localization/Features/Capture.xcstrings
+++ b/Snapzy/Resources/Localization/Features/Capture.xcstrings
@@ -1,16713 +1,16713 @@
 {
-  "sourceLanguage" : "en",
-  "strings" : {
-    "after-capture.accessibility-label" : {
-      "extractionState" : "stale",
-      "localizations" : {
-        "de" : {
-          "stringUnit" : {
-            "state" : "translated",
-            "value" : "%@ für %@"
+  "sourceLanguage": "en",
+  "strings": {
+    "after-capture.accessibility-label": {
+      "extractionState": "stale",
+      "localizations": {
+        "de": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "%@ für %@"
           }
         },
-        "en" : {
-          "stringUnit" : {
-            "state" : "translated",
-            "value" : "%@ for %@"
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "%@ for %@"
           }
         },
-        "es" : {
-          "stringUnit" : {
-            "state" : "translated",
-            "value" : "%@ para %@"
+        "es": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "%@ para %@"
           }
         },
-        "fr" : {
-          "stringUnit" : {
-            "state" : "translated",
-            "value" : "%@ pour %@"
+        "fr": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "%@ pour %@"
           }
         },
-        "ja" : {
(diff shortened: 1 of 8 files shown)


duongductrong:
Hi @omarshahine,

Thanks for the great suggestion! I did some quick manual testing and found that it works flawlessly for a single link. However, when processing a larger text with multiple links via OCR, it seems to fail.

As a user, I think it would be great if the tool could show all the detected links. What are your thoughts on this? I'd love to hear your feedback

omarshahine:
Thanks for testing it! You're right — I had scoped it down in 9110fa9 so the prompt only fired when the entire capture was a single URL. I've reverted that in bfe14b9: the prompt now surfaces all web links detected anywhere in the captured text again (deduped, capped at 3 rows to keep the panel compact, each row clickable). PR description updated to match. Let me know if you'd like the cap raised or removed.

Snapzy/Resources/Localization/Features/Capture.xcstrings
duongductrong, line 1: Hey @omarshahine,

Please make sure the agent doesn't format this file, it seems like running the format script and format over 11k LoCs

duongductrong (changes requested):

Snapzy/Resources/Localization/Features/Capture.xcstrings
omarshahine: Good catch, sorry about that — the edit script re-serialized the whole catalog with Xcode's " : " separator style instead of preserving yours. Fixed in d7a2bec: the file is now byte-identical to master except for the five inserted keys (+325 lines, 0 deletions).

omarshahine:
Heads-up on the red build check: both failing tests are unrelated to this PR. PersistedCombineSessionTests fails identically on master's own CI run (run 29200945865) — that suite doesn't exist on this branch. AreaSelectionMultiMonitorReconciliationTests passes locally and looks runner-flaky. The merge ref compiles cleanly and all new OCRLinkDetectorTests pass. Happy to rebase once master is green.

duongductrong (approved):
Hey @omarshahine , LGTM! Thank you for your effort

PR #374, opened 2026-07-15
fix: make scrolling capture reliable on mixed-scale external displays

### LINKED ISSUE #371: [Bug]: Scrolling capture fails on mixed-scale external displays and ignores the first selection drag
### Description

Scrolling capture does not work correctly when a 1x external display is connected to a Mac with a 2x built-in display.
The initial capture frame and subsequent live stream frames use different pixel dimensions, causing the stitcher to reject the scrolling frames.

Additionally:
- An area selection overlay may remain visible after a capture.
- The first drag when selecting a scrolling capture region may be ignored completely.
- The region can only be selected on the second drag.

Tested with:
- Built-in display: 1512×982 @2x
- External USB-C display: 1920×1080 @1x
- macOS 26.5.2

### Steps to Reproduce

1. Connect a 1x external display to a Mac with a 2x built-in display.
2. Launch Snapzy.
3. Trigger scrolling capture.
4. Move the pointer to the external display.
5. Drag to select a scrolling capture region.
6. Observe that the first drag may not display a selection rectangle.
7. Drag the region a second time.
8. Scroll the selected content.
9. Save the capture.
10. Observe that the result may contain only the initial visible frame instead of the full scrolling content.
11. Trigger a normal area capture and observe that the selection overlay may remain visible afterward.

### Expected Behavior

- The first drag should immediately create a selection rectangle.
- Scrolling capture should stitch frames correctly on both 1x and 2x displays.
- Mixed-scale display configurations should not produce mismatched frame dimensions.
- Selection overlays should be dismissed after capture or cancellation.

### Actual Behavior

- The first drag may produce no selection rectangle.
- A second drag is required to select the scrolling capture region.
- The initial frame is generated at the 2x session output scale.
- Subsequent ScreenCaptureKit frames remain at the external display’s native 1x scale.
- Example frame dimensions:
  - Initial frame: `2562×1680`
  - Live stream frame: `1281×840`
- The stitcher rejects the mismatched frames.
- Session metrics report `appended=0` with repeated `ignored-alignment-failed` outcomes.
- A selection overlay may remain visible after capture.

### Screenshots / Recordings

_No response_

### Snapzy Version

1.29.1 (Build 136)

### macOS Version

### Installation Method

None

### Additional Context

Three related causes were identified:
1. **Mixed display scale factors**

   The initial still frame used the session’s 2x output scale, while live ScreenCaptureKit frames remained at the external display’s native 1x scale.

2. **Selection dismissal state leakage**

   A live area selection could set `dismissesAfterSelection` to `false`. Cancellation did not restore its default value, allowing the setting to affect later capture sessions.

3. **Synthetic mouse event race**

   When pointer ownership moved to another display, Snapzy posted a synthetic mouse-down and mouse-up to refresh cursor ownership. If those events were not routed back to the selection overlay, the “ignore next mouse event” flags remained set and discarded the user’s first real drag.

motoish:
## Summary
Fix scrolling capture failures on mixed-scale display configurations, particularly when capturing from a 1x external display connected to a Mac with a 2x built-in display.

This also fixes two related area-selection issues:
- The selection overlay could remain visible after capture or cancellation.
- The first drag used to select a scrolling capture region could be ignored.

## Problem
The initial still frame was generated at the scrolling session's output scale, while subsequent ScreenCaptureKit frames remained at the external display's native scale.

For example:
- Initial frame: `2562×1680`
- Live stream frame: `1281×840`

Because the stitcher requires consistent frame dimensions, it rejected every live frame with `ignored-alignment-failed`, resulting in `appended=0`.

Area selection also had two lifecycle problems:
1. `dismissesAfterSelection` could remain disabled after a cancelled live selection and affect later capture sessions.
2. Cross-display pointer promotion injected synthetic mouse events and marked the next mouse-down and mouse-up for suppression. If the synthetic events were not routed back to the overlay, the user's first real drag was discarded.

## Testing
An unsigned Debug build completed successfully with:
```sh
xcodebuild \
  -project Snapzy.xcodeproj \
  -scheme Snapzy \
  -configuration Debug \
  -destination 'platform=macOS' \
  CODE_SIGN_IDENTITY= \
  CODE_SIGNING_REQUIRED=NO \
  CODE_SIGNING_ALLOWED=NO \
  build
```

Fixes #371

### DIFF
--- a/Snapzy/Services/Capture/AreaSelectionWindow.swift
+++ b/Snapzy/Services/Capture/AreaSelectionWindow.swift
@@ -1041,25 +1041,6 @@ final class AreaSelectionController: NSObject {
     // Invalidate cursor rects before the next event
     window.overlayView.refreshCursor()
     
-    // WindowServer fundamentally refuses to assign native cursor ownership to a newly-key window
-    // of an inactive application unless a hardware click occurs. We bypass this by posting a
-    // synthetic hardware click via CGEvent exactly at the current mouse position.
-    window.overlayView.ignoreNextSyntheticClick = true
-    window.overlayView.ignoreNextSyntheticMouseUp = true
-    
-    if let source = CGEventSource(stateID: .hidSystemState),
-       let currentEvent = CGEvent(source: nil) {
-      let cgMousePos = currentEvent.location
-      
-      let mouseDown = CGEvent(mouseEventSource: source, mouseType: .leftMouseDown, mouseCursorPosition: cgMousePos, mouseButton: .left)
-      let mouseUp = CGEvent(mouseEventSource: source, mouseType: .leftMouseUp, mouseCursorPosition: cgMousePos, mouseButton: .left)
-      
-      DispatchQueue.main.asyncAfter(deadline: .now() + 0.05) {
-        mouseDown?.post(tap: .cghidEventTap)
-        mouseUp?.post(tap: .cghidEventTap)
-      }
-    }
-
     DiagnosticLogger.shared.log(
       .debug,
       .capture,
@@ -1075,6 +1056,7 @@ final class AreaSelectionController: NSObject {
 
   private func resetCallbacks() {
     isPresenting = false
+    dismissesAfterSelection = true
     stopPointerTracking()
     lumaRecapturingTask?.cancel()
     lumaRecapturingTask = nil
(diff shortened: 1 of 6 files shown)


duongductrong (approved):
lgtm! thanks @motoish

PR #377, opened 2026-07-16
fix(annotate): restore markup on combined images and enhance text editing

menmer0859:
## Summary

This PR addresses two annotation issues found during beta testing.

### 1. Mark every image in a combined image

- Fixes the issue where the second and later images in a combined image intercepted input, preventing users from placing markup on them.
- Markup tools now operate on the combined canvas, so every participating image can be marked as expected.

I am sorry this escaped the verification of the previous release and caused confusing behavior in the Combine Images workflow.

### 2. Enhance the text editing experience

- Adds clearer text presentation modes, natural text sizing, and canvas-aware starting text sizes.
- Improves text callout behavior so its pointer follows the target continuously around the label and remains visually integrated with the rounded text background.
- When the pointer target moves inside the label, the callout is hidden and the label remains a normal rounded rectangle.

The text callout interaction and visual direction are inspired by [Shottr](https://shottr.cc/). This is a Snapzy-native implementation adapted to Snapzy's annotation workflow.

## Validation

- `AnnotateImageOpsTests`
- `AnnotateTextEditingTests`
- `AnnotateViewportUIStateTests`

All targeted tests pass on macOS.


duongductrong (approved):
lgtm

PR #429, opened 2026-07-27
feat: Add default and remembered annotation tools

### LINKED ISSUE #427: [Feature]: Take screenshots and annotate. I hope users can choose the default tool. Now it is the first one every time, but the first one is not commonly used.
### Problem or Motivation

no

### Proposed Solution

1

### Alternatives Considered

1

### Feature Area

Screenshots

### Mockups / References

_No response_

### Additional Context

_No response_

lcopilot:
## Problem

Screenshot annotation always starts with the Selection tool. Users who primarily annotate with arrows, rectangles, text, or another drawing tool must switch tools after every capture.

Closes #427.

### Annotation preferences

Two settings were added under Settings > Annotate:

- Default Tool selects the tool used when a screenshot annotation session begins.
- Remember Last Tool starts the next session with the last tool explicitly chosen by the user.

The settings are mutually exclusive in the UI. When Remember Last Tool is enabled, the Default Tool row is disabled and visually dimmed. Its stored value is preserved and remains available as a fallback.

Only tools supported by the inline screenshot annotation toolbar are shown. Crop and Mockup are intentionally excluded.

### Startup behavior

| Remember Last Tool | Stored history | Tool used for a new session |
| --- | --- | --- |
| Off | Any value | Configured Default Tool |
| On | Valid supported tool | Last explicitly selected tool |
| On | Missing, invalid, or unsupported | Configured Default Tool |
| Any | Invalid or unsupported default | Selection |

The default remains Selection, so existing users see no behavior change unless they update these settings.

### What counts as the last tool

The remembered value is updated only when the user explicitly activates a tool in an inline screenshot annotation session through:

- the inline annotation toolbar
- an annotation keyboard shortcut

Internal state changes do not update the remembered value. For example, clicking an existing annotation may temporarily change `selectedTool` so its properties can be edited, but that does not replace the user's remembered startup tool.

Tool choices made in the full annotation editor are not persisted by this preference.

### Persistence and configuration

The following TOML fields were added:

```toml
[annotate]
default_tool = "selection"
remember_last_tool = false
```

Both fields support import, export, default-document generation, validation, and configuration synchronization.

The transient last-used value is stored only in UserDefaults as `annotate.lastUsedTool`. It is intentionally not exported, so moving a configuration file to another Mac does not transfer incidental editing history.

### Localization and documentation

- added localized setting titles and descriptions for all 10 supported locales
- updated the annotation and configuration documentation
- documented the startup precedence and local-only history behavior

## Impact

- Runtime behavior changes only for new inline screenshot annotation sessions.
- Existing captures and annotations are not migrated or modified.
- Existing users continue to start with Selection because remembering is off by default.
- The full annotation editor keeps its existing startup and persistence behavior.
- Capture, recording, export, and annotation rendering behavior are unchanged.
- Invalid or stale preference values cannot activate Crop, Mockup, or an unknown tool in the inline overlay.

## Implementation notes

- `AnnotateToolPreference` centralizes validation, fallback selection, and last-tool persistence.
- `AnnotateState` exposes an explicit tool-activation callback.
- Toolbar and keyboard activation paths use `activateTool`, while internal `selectedTool` assignments remain silent.
- `InlineAreaAnnotateSession` resolves the startup tool after loading the captured image and records explicit activation through the callback.
- The callback captures UserDefaults rather than the session, so it does not introduce a retain cycle.

## Performance considerations

- Startup tool resolution is a constant-time preference lookup and enum validation.
- No additional work is added to capture, rendering, or export loops.
- When Remember Last Tool is enabled, each explicit tool activation writes one scalar UserDefaults value. That notification can schedule the existing configuration sync, whose 1.2-second debounce coalesces rapid changes.
- No persistence occurs during drawing, pointer movement, or internal selection changes.
- The added preference values are small scalar UserDefaults entries; no annotation payloads or images are persisted.

## Testing

Passed:

- localization catalog verification: 1450 keys across 10 locales, with no missing or extra translations
- `PreferencesCoreTests`
- `AnnotateCoreTests`
- `SnapzyConfigurationImporterTests`
- `SnapzyConfigurationServiceTests`
- final targeted run: 147 tests, 0 failures
- manual smoke test covering configured defaults and remembered tools
- `git diff --check`

A local full-suite attempt completed 1,219 tests successfully and skipped 1. The remaining 43 tests crashed in unrelated test-host areas such as cursor control, smart-element capture, audio metering, and sandbox migration; one existing activation-policy assertion also failed. No modified test suite failed.

Standard optimized Release compilation also hits a Swift 6.2.4 `SILPerformanceInliner` compiler crash under Xcode 26.3. Debug and non-optimized Release configurations compile successfully.

## PR history

This replaces #428, which GitHub automatically closed when the cross-repository head branch was renamed to follow the project's `feat/**` branch convention.


duongductrong (approved):
LGTM 👍

Reviewed the full diff and verified against the codebase:

- **Correct wiring**: `activateTool` already existed and is used by the inline toolbar (`InlineAreaAnnotateWindow.swift:1501`); the new `onToolActivated` hook is only attached in `InlineAreaAnnotateSession`, so full-editor tool choices are not persisted — matches the PR description.
- **Keyboard shortcut path**: `DrawingCanvasNSView.keyDown` is shared by both editors (`CanvasDrawingView` used in `AnnotateCanvasView` and `InlineAreaAnnotateWindow`), so shortcut-driven tool switches are remembered only in inline sessions. Internal `selectedTool` assignments (e.g. clicking an existing annotation) correctly bypass the hook.
- **Validation**: `AnnotateToolPreference` rejects unsupported/invalid stored values (crop, mockup, garbage strings) and falls back to Selection — covered by new unit tests.
- **Config export/import** round-trips the two new keys, docs updated (`ANNOTATE.md`, `CONFIGURATION.md`), and localization added for all 10 locales.
- Existing users see no behavior change (default stays Selection). CI green (build + swiftformat).

Nice work — safe to merge.

PR #430, opened 2026-07-27
feat: Extend live capture passthrough to menu bar popovers

tukuyomil032:
## Problem

PR #416 introduced event-tap passthrough for live capture so an already-visible hover UI is not dismissed when Snapzy begins selection. Menu-bar popovers remain a separate case: some third-party status-item popovers close while the selection panels are being presented, then vanish from the later asynchronous window query. Some can still be enumerated by WindowServer but cannot be rendered independently through Core Graphics.

## Solution

Extend the live-capture handling introduced in #416 to already-open, third-party menu-bar popovers.

- Identify only visible, nonzero-layer popovers anchored to the menu-bar edge; keep ordinary layer-0 application windows unchanged and exclude Snapzy, Dock, Control Center, and Notification Center.
- Snapshot the containing display before the selection overlay appears, crop the eligible popover, and merge that retained candidate ahead of the normal live window list.
- Preserve the normal capture order: ScreenCaptureKit when the exact window remains shareable, retained image when it has disappeared, Core Graphics single-window capture for a live eligible popover, then the existing area fallback.
- Restore a vanished retained popover beneath the existing application-selection dim cutout. A short, session-scoped WindowServer check prevents duplicate rendering when the source popover remains visible.
- Apply a transparent rounded-corner mask to retained popover images so desktop pixels do not leak into exported PNG corners.

## Scope

This intentionally covers only third-party popovers/dropdowns that were already open and anchored from the menu bar when capture begins. Quick Look, ordinary in-app popovers, and dedicated macOS system UI remain out of scope.

## Validation

- `xcodebuild test -project Snapzy.xcodeproj -scheme Snapzy -configuration Debug CODE_SIGNING_ALLOWED=NO`
- `WindowCaptureSelectionPolicyTests`: 11 tests, 0 failures
- `git diff --check`
- Manual QA with CodexBar: open the Claude status-item popover, start area capture, switch to Application Capture, select the popover, and save it after the original source UI closes.
- Verified retained PNG corners are transparent rather than carrying the underlying desktop pixels.

## Notes

The selection overlay itself uses `sharingType = .none`, so its retained-image restoration is intentionally not visible in a screen recording. The saved capture and the live selection UI are the relevant verification surfaces.


duongductrong (approved):
LGTM 👍

Reviewed the full diff against master in context. Verdict: merge.

**Verified**
- Ordering is correct: `captureImmediateMenuBarPopoverCaptures` runs synchronously before any selection panel is presented (`CaptureViewModel.swift`), and the session seeding → `activatePooledWindows` → visibility refresh → snapshot merge sequence in `AreaSelectionController` is sound (session-ID + `isPresenting` guards, task cancelled on both teardown paths).
- Retained candidates merge ahead of the live layer-0 list with windowID dedupe, so hit-testing prefers the popover; live query still excludes nonzero layers — scope stays limited to UI visible at capture start, as documented.
- Capture path resolution (SCK → retained image → CG single-window → area fallback) matches the PR description; `.areaFallback` preserves the previous log + behavior for normal windows.
- Retained layers sit above `snapshotLayer` and below `dimLayer`, so the existing application-mode cutout reveals them; hidden while the source window is still on screen — no double rendering.
- Coordinate math (AppKit↔Quartz Y-flip, retina scale, offset-display clamping) is covered by the new unit tests; I recomputed the expected rects and they check out. The masked crop is a fresh bitmap, so the full-display snapshot is freed after masking.
- `refreshWindowPool` only calls `configureSessionWindow` for newly attached displays, so the `setRetainedMenuBarPopoverCaptures` state reset can't hide a restored popover on an existing display mid-session.
- New files are picked up by the synchronized folder groups (no pbxproj change needed); CI test suite + Release build + swiftformat all green; `CGDisplayCreateImage` usage matches the existing fast path on master. Docs (CAPTURE.md, STRUCTURE.md) and README updated.

**Non-blocking nits**
- `rawWindowInfoList()` duplicates the window-info parsing already inline in `prepareSnapshot` — worth sharing later.
- If SCK resolution succeeds but the capture itself throws for a popover target, the fallback goes to area capture rather than the retained image — matches prior behavior, just noting.

Nice work — clean scoping, good test coverage, thorough docs.

tukuyomil032:
Thank you for creating such a great app! 
Snapzy is hands-down the easiest-to-use, most feature-rich screenshot app I’ve ever tried, and I really love its sleek UI!

PR #432, opened 2026-07-28
Fix ja localization

kawarimidoll:
## Summary

Fixes the Japanese (`ja`) localization, makes the Japanese UI read more naturally. No other locale, key, or metadata is changed.

Example: Onboarding message

| Current master | After this PR |
|---|---|
| SCR-20260728-sqlz | SCR-20260727-tkfs |

One class of problem per commit:

| Commit | Examples |
|---|---|
| repair broken strings | text leaked in from neighbouring entries, a `%@` the source string does not have, a Simplified Chinese character, zero-width spaces |
| correct mistranslated terms | `Stroke` became the medical condition (the bug #321 fixed for `es`/`fr`), `Free` became "free of charge" (#367 for Chinese), `Granted` became the adverb "indeed" |
| screen-recording wording | the recording toolbar used the word for recording *audio* (#296 for Chinese); `Restart` and `Resume` had collapsed onto one phrase |
| product names | "Snapzy" transliterated into invented katakana, including a save path that does not exist on disk |
| command wording | buttons phrased as polite requests: "Check for Updates..." read as "please go and check for updates" |
| spelling and spacing | spaces inside compound words, terms spelled two ways at once |

## Verification

- `swift tools/localization/CatalogTool.swift verify` → passed, `missing=0`, `extra=0`
- `xcodebuild ... -configuration Debug build` → succeeded
- Checked programmatically that only `ja` values differ from `master`, and that every
  changed value keeps the format specifiers of its `en` counterpart


duongductrong (approved):
lgtm

PR #441, opened 2026-07-31
chore: update changelog shell

motoish:
## Summary
- Drop the blank lines between the file preamble and the first `## [x.y.z]` heading before inserting a new entry in `scripts/update-changelog.sh`.
- Add a small `print_without_trailing_blanks` awk helper and reuse it for both the "existing entries" and "no entries yet" paths.
- Normalize entry output with `printf '%s\n'` so spacing is exactly one blank line between the preamble and the newest entry.

## Why
`update-changelog.sh` kept everything above the first version heading with `head -n $((HEADER_END - 1))` and then added one more `echo ""`, so every release added one extra blank line. `CHANGELOG.md` has drifted to 105 blank lines between the preamble and the first entry (`## [1.31.0-beta.1]` now starts at line 111), which also makes the diff of every release commit noisy.

The stale blank lines are intentionally not stripped in this PR — the next release run collapses them to a single blank line, so no manual cleanup commit is needed.

### DIFF
--- a/scripts/update-changelog.sh
+++ b/scripts/update-changelog.sh
@@ -35,27 +35,39 @@ NEW_ENTRY="## [${VERSION}] - ${DATE}
 
 ${CONTENT}"
 
-# Find the line number of the first "## " heading (previous entry) or end of file
-# We insert after the file header (first 5 lines: title + blank + description + blank + format line)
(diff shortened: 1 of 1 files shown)


duongductrong (approved):
lgtm

PR #455, opened 2026-08-02
feat(annotate): add Shift drawing constraints

tukuyomil032:
## Problem

Annotation tools currently draw freely from the initial pointer position, making it difficult to create precise squares, circles, and angle-aligned lines or arrows. The final annotation must also match the constrained preview when the raw mouse-up position differs from the displayed endpoint.

## Solution

- Add `AnnotationDragConstraint.constrainedEndPoint(tool:arrowStyle:start:end:shiftHeld:)` as a pure constraint helper.
- While Shift is held:
  - Rectangle, Filled Rectangle, and Oval use equal width and height while preserving the drag direction, including axis-aligned drags.
  - Line and Straight Arrow snap to the nearest 45-degree increment while preserving drag distance.
  - Curved arrows and unsupported tools remain unconstrained.
- Leave input unchanged when Shift is not held.
- Apply the helper to non-freehand previews in the shared `DrawingCanvasNSView`, covering both the full Annotate editor and inline area annotation.
- Commit non-freehand annotations using the final preview endpoint so the saved result matches the preview.
- Keep Pencil and Highlighter gestures freehand.

## Scope

- Add focused model tests for shape constraints, 45-degree snapping, unsupported tools, and unconstrained input.
- Add regression coverage for mouse-up coordinates differing from the preview endpoint, including horizontal, vertical, and diagonal line snaps.
- Document the Shift shortcut in `docs/ANNOTATE.md`.
- Do not change `AnnotationFactory`, persistence or session formats, or existing resize interactions.

## Validation

Completed:

- Full test suite passed: 1,557 passed, 0 failed, 3 skipped.

```sh
xcodebuild -quiet test \
  -project Snapzy.xcodeproj \
  -scheme Snapzy \
  -configuration Debug \
  -derivedDataPath /private/tmp/snapzy-upstream-rebuild-derived-data \
  CODE_SIGN_IDENTITY='' \
  CODE_SIGNING_REQUIRED=NO \
  CODE_SIGNING_ALLOWED=NO
```

- `git diff --check` passed.

The three skipped checks are opt-in: a real PP-OCR model integration, a real microphone integration, and a PNG-only visual snapshot renderer.

Manual QA completed:

- [x] In the full Annotate editor, held Shift while dragging Rectangle, Filled Rectangle, and Oval in every direction; confirmed equal width and height.
- [x] In the full Annotate editor, held Shift while dragging Line and Straight Arrow; confirmed snapping to the nearest 45-degree increment.
- [x] Confirmed the committed annotation matches the constrained preview after mouse-up.
- [x] Confirmed Curved Arrow, Pencil, and Highlighter behavior remains unconstrained.
- [x] Repeated the applicable checks in inline area annotation.
- [x] Confirmed drawing without Shift preserves the existing behavior.
- [x] Confirmed existing annotation resize interactions are unchanged.

## Notes

- No screenshots are included.


duongductrong (changes requested):
This review has been superseded by the corrected review submitted below. Please use the newer review for the actionable feedback.

duongductrong (changes requested):
## Request changes

### Shift-constrained geometry can escape the active canvas bounds

**Location:** `Snapzy/Features/Annotate/Components/AnnotateCanvasDrawingView.swift`, `mouseDragged(with:)`, lines 901–908.

`interactionPoint(from:)` already clamps the raw pointer to the active drawing bounds. The new code then transforms that clamped point with `AnnotationDragConstraint.constrainedEndPoint(...)` and stores the result in `currentPath` without applying the bounds again. A valid pointer position can therefore produce an out-of-bounds annotation.

For example, on a 400-point-wide canvas, starting at `(395, 100)` and Shift-dragging vertically to `(395, 140)` makes `squareEndPoint` return `(435, 140)`. The preview and committed rectangle then extend 35 points beyond the right edge. The same issue can occur for snapped lines and straight arrows.

Please update the call site/helper so the constrained endpoint remains within the same bounds used by `interactionPoint(from:)` (`activeAnnotationBounds`, or `effectiveContentBounds` in combine mode). Making the constraint calculation bounds-aware would preserve the intended square/45-degree behavior as much as possible; clamping the final endpoint is the minimum safety fix, although it may distort that invariant at the edge.

Please also add regression coverage for the relevant canvas edges in the shared full-editor and inline-annotation path. The current CI and focused tests cover interior drags, but not this boundary case.

tukuyomil032:
Thanks for the detailed review — I implemented the requested bounds fix in be3b77d.

- Made `AnnotationDragConstraint` bounds-aware, so constrained shapes preserve equal dimensions and snapped lines/arrows preserve their angle while shortening to the available distance at the active canvas edge.
- Reused the same active drawing bounds for raw pointer clamping and constrained previews (`activeAnnotationBounds`, or `effectiveContentBounds` in Combine mode).
- Added regression coverage for edge-constrained shapes, lines, straight arrows, Combine with a crop, and zero-capacity drags that must not create invisible annotations.
- Also fixed stationary Shift changes: pressing or releasing Shift without moving the pointer now updates the preview immediately, and mouse-up commits that latest preview. The canvas explicitly becomes first responder, with an NSWindow event-delivery regression test.

Validation:

- `AnnotationDragConstraintTests` and `AnnotateCoreTests` pass.
- `git diff --check` passes.
- Manual QA is complete in both the full Annotate editor and inline area annotation.

I also ran the full suite. Its post-change runs were interrupted nondeterministically by pre-existing crashes in unrelated AreaSelection, HistoryThumbnail, and ImageDrop tests; the reported AreaSelection case passes in isolation and does not reference the annotation canvas changes.

duongductrong (approved):
lgtm! please ignore conflicts, will fix soon, thanks for your effort @tukuyomil032

PR #466, opened 2026-08-05
feat: enlarge screenshot magnifier and add pixel grid, crosshair, color picker

### LINKED ISSUE #426: [Feature]: Make the Screenshot Magnifier Larger and Add a Crosshair
### Problem or Motivation

no

### Proposed Solution

Image

The current magnified preview in the screenshot tool is too small to be genuinely useful. It is difficult to see individual pixels or determine whether the selection is aligned precisely with the edge of a window, icon, or text.
Please consider making the preview larger, similar to the reference image, and adding a clear crosshair to indicate the exact pixel being selected. Displaying the current selection dimensions or cursor coordinates would also make precise screenshots much easier.
The current implementation makes fine adjustments difficult and often requires several attempts to get the desired result. A larger preview and crosshair would significantly improve accuracy and usability.

### Alternatives Considered

Currently still using CleanShot X

### Feature Area

Screenshots

### Mockups / References

Image

### Additional Context

_No response_

lcopilot:
## Summary

Resolves #426.

- Grows the area-selection magnifier preview (160×160pt, up from 130×130) and adds a pixel grid, a gapped crosshair on the target pixel, and a color-picker panel below it showing cursor coordinates, a color swatch, and the hex value.
- Fixes a pre-existing precision bug: the highlighted target-pixel box didn't account for the backdrop's own resolution, so on Retina displays it covered more than one source pixel instead of exactly one. The grid and crosshair now line up exactly with what's actually rendered.
- The pixel grid appears once ⌘+scroll zoom passes a fixed threshold (zoom ≥ 6), independent of display resolution, so the same amount of scrolling always crosses it regardless of the screen.
- Adds a "C" shortcut to copy the sampled color to the clipboard, with a brief "Copied" confirmation, shown as a key-cap badge next to the copy hint.
- Adds a "show magnifier by default" preference (Preferences → Screenshot → Magnifier Zoom, off by default so existing behavior is unchanged unless opted in) so the magnifier can start already active when a selection session begins, instead of requiring ⌘+scroll to activate it first. Wired through configuration import/export.
- Brings the same magnifier to screenshot-and-annotate (⌘⇧7), which previously had none — that flow is a separate SwiftUI-based selection UI, so the shared AppKit/CALayer `AreaSelectionMagnifier` is bridged in via a new `InlineAreaMagnifierHostView` (`NSViewRepresentable`) rather than reimplementing the rendering a second time. Both entry points now share one implementation and behave identically.
- New copy hint / feedback strings are localized across all 10 supported languages.

## Test plan

- [x] `AreaSelectionOverlayMagnifierTests` / `AreaSelectionOverlayMagnifierLayoutTests` (new + updated cases: sizing/flip positioning, pixel grid visibility threshold — including a resolution-independence regression test, crosshair, color panel content, copy-to-clipboard, "C" key handling)
- [x] `InlineAreaAnnotateSessionTests` (new cases for the `.copyMagnifierColor` key action during `.selecting` vs `.annotating`)
- [x] `SnapzyConfigurationServiceTests` / `SnapzyConfigurationImporterTests` (new preference round-trips through export/import)
- [x] Manually exercised both ⌘⇧4 (plain area screenshot) and ⌘⇧7 (screenshot-and-annotate) selection flows: magnifier sizing, grid/crosshair alignment, color panel, ⌘+scroll zoom, "C" copy, and the "show magnifier by default" preference toggle
- [ ] Reviewer: please sanity-check the non-English key-cap hint translations (`screen-capture.magnifier-copy-color-hint-prefix` / `-suffix` in `Capture.xcstrings`) — translated without native review


duongductrong:
Hi @lcopilot,

Thank you for your work on this PR!

After reviewing the changes from a user experience perspective, I have a few observations and recommendations for improvement:

1. **Magnifier Contrast on Light Backgrounds**
   When the magnifier moves over white or light-colored backgrounds, its boundaries blend into the content, reducing visibility compared to the previous behavior.
   - **Recommendation**: Apply a subtle drop shadow or contrast border around the magnifier boundary to improve visual distinction across bright surfaces.

2. **Cursor Position Instability During Mouse Wheel Scroll**
   When scrolling with the mouse wheel to adjust magnifier zoom, the cursor occasionally jumps abruptly between coordinates.
   - **Recommendation**: Please verify the `scrollWheel` event handling and coordinate translation logic to ensure smooth and stable positioning during scroll-based zooming.

3. **Redundant Coordinate Displays**
   Both the crosshair overlay and the color picker component currently display position coordinates `(X, Y)`.
   - **Recommendation**: Remove the coordinates from the color picker UI and retain the crosshair as the primary display for coordinate information.

4. **Color Picker Screen Footprint & Toggle Option**
   The color picker container currently consumes a relatively large area of the capture interface.
   - **Recommendation**:
     - Make the color picker component more compact (removing redundant coordinate labels will help streamline the design).
     - Provide a user preference toggle to enable or disable the color picker overlay based on individual workflow needs.

Please let me know your thoughts on these points. Thank you again for your contribution!

duongductrong:
Thanks so much for clarifying, @lcopilot ! That makes complete sense.

Based on your feedback, here is the proposed update:

1. **Restore original coordinate bubble untouched**: Revert all changes to the original cursor coordinate bubble so it stays strictly in its original format (`X, Y`) and appearance. Modifying the crosshair coordinate bubble is outside the scope of this color picker and magnifier PR.
2. **Remove coordinate row from color picker panel**: Remove the `(x, y)` coordinate line from the color picker panel completely to keep the panel compact and eliminate duplicate coordinate displays.

Please let me know if this proposed direction aligns with your expectations. If so, I'll update the PR code accordingly. Thank you!

hungdev:
nice features, hopefully these features will be merged soon.

duongductrong:
Hi @lcopilot ,

Thanks for your work on this PR and for sharing your feedback!

To give a bit of context on our design vision of the crosshair and coordinator: in earlier versions, Snapzy did use a darker overlay with more prominent handle indicators. However, our current product direction is to stay as close as possible to the built-in macOS screenshot tool (Apple native style). This makes the experience feel familiar and seamless for macOS users.

If the native design feels less visible for some users, a good long-term solution would be adding an **Appearance** tab in Settings/Preferences, allowing users to customize the UI overlay and handle visibility based on their preference.

Regarding the color picker, I now understand your reasoning behind doubling the indicator—it’s a very valid point that I had overlooked, and I’m totally fine with keeping that change for the color picker context.

To summarize:
- We can keep the handle indicators for the **color picker** as designed in your PR.
- However, for the **main bubble selection handles**, I’d appreciate it if you could revert them back to Snapzy’s existing (near-native) version for the reasons mentioned above.

Let me know your thoughts!

duongductrong:
Thank you for your effort @lcopilot, LGTM!

duongductrong (approved):

PR #472, opened 2026-08-07
feat: reduce friction in window/element capture — auto-detect under cursor, layered smart element detection

lcopilot:
Hi @duongductrong,

Thanks again for maintaining Snapzy and for the thoughtful review on #466 — really enjoyed working through that with you. Wanted to open this one for your consideration as well, and I'd genuinely welcome your thoughts on the approach before we take it further.

## Why

Snapzy already has a lot of powerful capability under the hood — window capture, and AX-based element-level capture via Smart Element Capture — but today they're mostly reached through separate, deliberate steps: pressing "A" mid-selection to switch into window mode, or remembering a dedicated shortcut (⌥⇧4) for element capture. For the single most common thing people want to do — "just grab this window" or "just grab this button" — that's a fair number of steps and a fair number of shortcuts to keep in your head before you get any benefit from the capability that's already there.

This PR doesn't add a new capture capability. It just removes the friction in front of two that already exist, by making them discoverable through hovering instead of through a shortcut or a mode switch:

- Hover during a normal area selection, and you see what you'd capture if you clicked — no "A" needed.
- If you want more precision than "the whole window," the same hover can also prefer a specific element (a button, a popup, something inside a web page) using the exact same engine Smart Element Capture already uses.
- Drag like normal, and it's just a manual region selection, exactly as before — nothing about the existing flow changes unless you opt in.

## What's in this PR

**1. "Automatically detect window under cursor" (off by default)** — for both plain area screenshot and screenshot-and-annotate:
- Hovering during selection highlights the window under the cursor.
- A plain click captures it directly.
- A drag past a small threshold (4pt) falls through to a normal manual region selection — so nothing about the existing drag-to-select muscle memory changes.

**2. "Also detect specific elements, not just windows" (off by default, requires #1)** — layers the existing Smart Element Capture engine on top of the same hover: when a specific element is found, it's highlighted instead of the whole window; when it isn't (e.g. an app with a sparse accessibility tree), it falls back to the window. Kept as its own toggle rather than folding it into #1, since detecting elements needs Accessibility permission and a slightly different mental model, and some users may only want the coarser, permission-free window behavior.

Both reuse existing infrastructure end-to-end — `WindowSelectionQueryService` (the same window enumeration the "A" toggle already uses) and `SmartElementQueryService` (the same engine behind the standalone Smart Element Capture shortcut) — so this isn't a second implementation of either, just a new, lower-friction way to reach them.

## A couple of implementation notes

- Click-vs-drag is disambiguated with a small movement threshold, mirrored identically across the AppKit-based plain area screenshot and the SwiftUI-based screenshot-and-annotate, so the two entry points behave the same way.
- While wiring this up, found and fixed a pre-existing staleness bug: a couple of code paths read `NSEvent.mouseLocation` / `CGEvent(source: nil)?.location` directly, which goes stale during live-passthrough capture (its event tap consumes mouse-moved events for its own tracking). It was there before this PR but only became obvious once a small, precisely-positioned element highlight made a frozen position visually obvious — the coarser whole-window highlight had been masking it. Fixed by deriving position from the already-live point each call site already has, rather than re-reading global state.
- Added unit tests for the click/drag disambiguation logic; manually verified hover/click/drag against both native macOS apps and Electron apps (to confirm the window-level fallback for apps with sparse accessibility trees).

## Questions for you

- Default-off felt like the safer choice for both toggles so existing habits aren't disrupted — but happy to hear if you'd rather see the window-level one on by default, given it's non-destructive (drag still works normally).
- Two dependent toggles vs. one combined toggle — open to collapsing these if you think that's a clearer story for users.

## Branch note

This branch includes #466 (still under review), since it builds on the shared magnifier component. Happy to rebase once that lands, or split this into an independent branch first if you'd rather review it in isolation — whatever's easier on your end.

Thanks for maintaining this project — happy to adjust anything here based on your take!


duongductrong:
Thank you for your contribution, @lcopilot ! 

I'm still working review on this PR, and I'll get back to you with an update soon. Thanks for your patience and for waiting!

duongductrong:
Hi @lcopilot, thanks for this! I did some hands-on testing and love the idea. Found 2 issues:

Snapzy_2026-08-20_21-28-16_581

**1. Cannot select an area by dragging**

- Enable: "Automatically detect window under cursor"
- Do: Open a capture, then drag to select a region
- Expected: The selection rectangle appears and the region is captured
- Actual: Nothing happens, no selection rectangle at all

**2. Cannot capture the whole window**

- Enable: "Also detect specific elements, not just windows"
- Do: Hover over a window and click
- Expected: The whole window gets captured
- Actual: The window is not captured, the wrong area gets captured instead

**Suggestion**

Maybe one smart toggle that figures out by itself whether you want an element, a window, or a custom area. The drag to select behavior should always keep working. It would also help to show a small hint of what will be captured before you click.

Happy to retest, the idea is really promising!

duongductrong (approved):
LGTM! 🎉 

Thank you @lcopilot for the well-structured PR and thorough consideration of existing flows:
- **Default OFF:** Keeping both settings off by default is definitely the right choice to guarantee zero disruption to existing muscle memory and 100% stability.
- **2 Dependent Toggles:** Having separate toggles for Window vs Element detection is very clean and avoids unnecessary Accessibility permission prompts for users who only want coarse window detection.
- **Live Passthrough Coordinate Fix:** Great catch on the stale mouse position during live-passthrough event interception.

All test suites (`SmartElementCaptureControllerTests`, `AreaSelectionAutoWindowDetectionTests`, `AreaSelectionLivePassthroughWindowDetectionTests`) pass cleanly.

PR #483, opened 2026-08-08
fix: lossless round-trip for function-key shortcuts in config export/import

YuriNachos:
`SnapzyConfigurationShortcutCodec.charToKeyCode` had no cases for function keys, so a binding like ⌘F5 did not round-trip through config export→import — the `F5` component was dropped/mangled on re-import.

## Fix
Add `F1`–`F20` cases (`kVK_F1`…`kVK_F20`) to `charToKeyCode`, so function-key shortcuts encode and decode back to the same key code. Single file, single concern.

## Test plan
`xcodebuild build` — the codec compiles. (The worker also produced a `SnapzyConfigurationShortcutCodecTests` regression suite; it is held back from this PR only because adding a new test file needs an `xcfilelist`/target-membership change — happy to follow up with it.)


duongductrong:
Thanks a lot for digging into this, @YuriNachos — the diagnosis is spot on: the exporter emits `F1`–`F20` labels but the importer's key parser had no cases for them, so function-key bindings were silently dropped on re-import.

I'm closing this one in favor of #484, which carries the identical fix plus a regression test suite covering the round-trip. One note for future contributions: this project uses Xcode synchronized folder groups, so new files under `SnapzyTests/` are picked up automatically — no xcfilelist or target-membership change needed. Your test suite was safe to include all along. 🙌

PR #486, opened 2026-08-09
fix(capture): pin DateFormatter to Gregorian locale for filename tokens

YuriNachos:
## Summary

Date tokens in capture filename templates (`{datetime}`, `{year}`, etc.) were resolved against the user's **system calendar** via a `DateFormatter` that did not pin its calendar. On a system using a Buddhist calendar (e.g. Thai `th_TH`), `yyyy` produced the **Buddhist year** (2569) instead of the Gregorian year (2026), so screenshots/recordings were named with the wrong year.

## Root cause

`CaptureOutputNaming.format(_:style:)` created a `DateFormatter` and set only `dateFormat`, leaving the calendar to whatever the current locale's calendar is. When that calendar is non-Gregorian (Buddhist, Hebrew, etc.), `DateFormatter` adopts it for token resolution.

## Fix

Pin the internal formatter's **calendar** to `Calendar(identifier: .gregorian)` — the narrower of the two options from the review. The year is always Gregorian (verified: `th_TH` + Gregorian calendar renders `2026 มกราคม` for `yyyy MMMM`), while `{monthName}` / `{monthShort}` keep following the system locale instead of being forced to English, so non-English users keep localized month names in their templates. File-format and template parsing are unchanged; only the date-token rendering path is affected.

## Tests

- `testYearToken_alwaysGregorian_notBuddhistCalendarYear` — `{year}` resolves to the Gregorian year (2026); includes a precondition guard proving `th_TH` would have rendered a different (Buddhist) year for the same date.
- `testDefaultTemplate_startsGregorianYear_notBuddhistYear` — the default template emits a Gregorian-year prefix, not a Buddhist-year prefix.
- `testGregorianCalendarOverride_keepsLocalizedMonthNames` — pins the calendar-override contract: a `th_TH` formatter with the Gregorian calendar yields the Gregorian year **with** the localized month name (`2026 มกราคม`).
- The month-name test helper now mirrors the production formatter (Gregorian calendar, locale left to the system) instead of using an unpinned `DateFormatter`, so its expectations match production output on any test-machine locale.

> Note: `Locale.current` is cached for the XCTest host process lifetime, so these tests cannot deterministically *fail pre-fix* on a Gregorian CI runner. They pin the Gregorian contract and will fail on a Buddhist-locale machine if the Gregorian calendar override is reverted.

## Verified

- `swift tools/localization/CatalogTool.swift verify` — unchanged (no string-table edits).
- `xcodebuild test -only-testing:CaptureOutputNamingTests` — 41 passed, 0 failed.

### DIFF
--- a/Snapzy/Services/Capture/CaptureOutputNaming.swift
+++ b/Snapzy/Services/Capture/CaptureOutputNaming.swift
@@ -212,11 +212,12 @@ enum CaptureOutputNaming {
 
   private static func format(_ date: Date, style: String) -> String {
     let formatter = DateFormatter()
-    // Force a Gregorian, POSIX-stable locale so date tokens always resolve to
-    // Gregorian values regardless of the user's system calendar. Without this,
-    // users on a Buddhist calendar (e.g. Thai `th_TH`) get the Buddhist year
-    // (e.g. 2569) instead of the Gregorian year (e.g. 2026) in filenames.
-    formatter.locale = Locale(identifier: "en_US_POSIX")
+    // Pin only the calendar so date tokens always resolve to Gregorian values
+    // regardless of the user's system calendar. Without this, users on a
+    // Buddhist calendar (e.g. Thai `th_TH`) get the Buddhist year (e.g. 2569)
+    // instead of the Gregorian year (e.g. 2026) in filenames. The locale is
+    // left alone so `{monthName}`/`{monthShort}` stay localized.
+    formatter.calendar = Calendar(identifier: .gregorian)
     formatter.dateFormat = style
     return formatter.string(from: date)
   }
--- a/Snapzy/Services/Capture/CaptureOutputNaming.swift
+++ b/Snapzy/Services/Capture/CaptureOutputNaming.swift
@@ -212,6 +212,11 @@ enum CaptureOutputNaming {
 
   private static func format(_ date: Date, style: String) -> String {
     let formatter = DateFormatter()
+    // Force a Gregorian, POSIX-stable locale so date tokens always resolve to
+    // Gregorian values regardless of the user's system calendar. Without this,
+    // users on a Buddhist calendar (e.g. Thai `th_TH`) get the Buddhist year
+    // (e.g. 2569) instead of the Gregorian year (e.g. 2026) in filenames.
+    formatter.locale = Locale(identifier: "en_US_POSIX")
     formatter.dateFormat = style
     return formatter.string(from: date)
   }
--- a/SnapzyTests/Services/Capture/CaptureOutputNamingTests.swift
+++ b/SnapzyTests/Services/Capture/CaptureOutputNamingTests.swift
@@ -500,8 +500,9 @@ final class CaptureOutputNamingTests: XCTestCase {
       defaults: defaults
     )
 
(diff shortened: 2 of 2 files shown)


duongductrong (changes requested):
Root cause + fix are correct — I reproduced it locally: under `th_TH` a bare `yyyy` renders `2569`, and pinning `en_US_POSIX` restores `2026`. Nice catch, and the tests are honest about not being able to fail pre-fix on a Gregorian runner.

Two things before merge.

### 1. `testResolveBaseName_monthNameTokens` now fails on non-English system locales (must fix)

The test helper at `SnapzyTests/Services/Capture/CaptureOutputNamingTests.swift:261` still uses an unpinned `DateFormatter`:

```swift
private static func format(_ date: Date, style: String) -> String {
  let formatter = DateFormatter()
  formatter.dateFormat = style
  return formatter.string(from: date)
}
```

`testResolveBaseName_monthNameTokens` compares production output against that helper. Before this PR both sides used the current locale, so they agreed. After the fix production emits `January` while the helper still emits the localized name, so the assertion at line 235 breaks on any non-English machine:

```
vi_VN  helper -> "tháng 1"   production -> "January"
fr_FR  helper -> "janvier"   production -> "January"
```

CI on `en_US` won't catch it. Please add `formatter.locale = Locale(identifier: "en_US_POSIX")` to the test helper so it mirrors production.

### 2. `{monthName}` / `{monthShort}` are now always English (needs a decision)

`en_US_POSIX` pins more than the calendar — it also pins the language. Non-English users who use `{monthName}` in a template go from `tháng 1` / `janvier` to `January`. If those tokens are used as subfolders, existing folder structures fork.

That may well be the intent (stable, ASCII-safe filenames), but it's a user-visible behavior change not mentioned in the PR description. If localized month names should be preserved, the narrower fix is a calendar override instead of a locale override — verified locally:

```swift
formatter.calendar = Calendar(identifier: .gregorian)
// th_TH -> "2026 มกราคม"  (Gregorian year, localized month)
```

Either choice is defensible. Please make it explicit — pick one, and if `en_US_POSIX` stays, call out the month-name change in the PR description and add a test pinning `{monthName} == "January"` so the contract is intentional rather than incidental.

Everything else looks good — `format` is the single choke point for all date tokens including the fallback path, so the one-line change covers `{datetime}`, `{date}`, `{time}`, `{ms}`, `{year_short}` and `fallbackName` too.

YuriNachos:
Thanks for the detailed review and for reproducing the locale behavior locally — went with the narrower `formatter.calendar = Calendar(identifier: .gregorian)`, so `{monthName}`/`{monthShort}` stay localized while the year is fixed; added `testGregorianCalendarOverride_keepsLocalizedMonthNames` pinning the `th_TH` case to `2026 มกราคม`. For the helper I mirrored the production formatter (Gregorian calendar, locale left to the system) rather than pinning `en_US_POSIX`: under the calendar-only fix production emits system-locale month names, so an `en_US_POSIX` helper would break the assertion the same way on non-English machines. Happy to switch to full `en_US_POSIX` plus an explicit `{monthName} == "January"` pin if you prefer the always-English contract.

duongductrong (approved):
lgtm, thank you so much @YuriNachos

PR #491, opened 2026-08-10
fix(cloud): percent-encode object-key path in AWS SigV4 presigned URLs

YuriNachos:
## Problem
`AWSV4Signer.presignURL` built its canonical URI straight from `url.path`, skipping the `addingPercentEncoding(.urlPathAllowed)` step that `sign()` applies. On Apple platforms `URL.path` returns the **decoded** path, so an S3/R2 object key containing a space (or any other non-`urlPathAllowed` character) — e.g. a custom filename template that yields `Snapzy 2024 shot.png` — leaked a raw character into:

- the **canonical request**, where AWS SigV4 requires percent-encoding, so the server reconstructed a different canonical URI and the signature did not match; and
- the **final URL string**, where the raw character made `URL(string:)` return `nil`, so `presignURL` threw `.signingFailed`.

The sibling `sign()` path was already correct; the encoding step was simply dropped when `presignURL` was written.

## Solution
- Mirror `sign()`'s encoding in `presignURL`: `url.path.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed) ?? url.path`, used for **both** the canonical request and the returned URL, so the two stay consistent.

## Scope
- Tests: added `testAWSV4SignerPresignURL_percentEncodesSpacesInObjectKey` (red before / green after) to `SnapzyTests/Services/Cloud/CloudCoreTests.swift`.
- NOT changed: `sign()` (already correct), canonical query-string/header handling, crypto helpers.

## Validation
- Localization catalog verify: OK (no UI strings touched).
- `xcodebuild build-for-testing` (scheme `Snapzy`, macOS destination): **TEST BUILD SUCCEEDED**. The targeted `xcodebuild test` run hits Snapzy's known headless "test runner hung before establishing connection" issue on this machine, so CI is the runner for the suite; the new test is deterministic (no OCR / no GUI / no network).
- `git diff --check` clean. Xcode 26.4.1 locally.

### DIFF
--- a/Snapzy/Services/Cloud/AWSV4Signer.swift
+++ b/Snapzy/Services/Cloud/AWSV4Signer.swift
@@ -120,7 +120,14 @@ enum AWSV4Signer {
     let credentialScope = "\(dateStamp)/\(region)/\(service)/aws4_request"
     let credential = "\(accessKey)/\(credentialScope)"
 
-    let canonicalURI = url.path.isEmpty ? "/" : url.path
+    // Percent-encode the path the same way `sign()` does: `url.path` returns the
+    // DECODED path on Apple platforms, so a key containing a space or other
+    // non-`urlPathAllowed` character would otherwise leak into both the canonical
+    // request (causing an AWS signature mismatch) and the final URL string (where
+    // the raw character makes `URL(string:)` return nil and the call throw).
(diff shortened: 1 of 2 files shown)


duongductrong (approved):
lgtm

PR #504, opened 2026-08-17
fix: keep space-to-move selection locked to the pointer

vxirau:
Follow-up to #294, which added space-to-move for area capture. Two defects there let the committed rect disagree with the one drawn on screen.

### Drift while moving

The move accumulated a per-event delta into the selection's own points. That holds only while a single source reports pointer movement, which was true when #294 landed. `updateManualSelection` is now reached from four:

- the local `NSEvent` drag monitor
- the global `NSEvent` drag monitor
- the overlay view's drag delegate
- the live-capture event tap, added later by #416 / #430

The monitors read `NSEvent.mouseLocation` live at callback time; the other two carry the recorded event location. The same physical movement therefore arrives with slightly different samples, and each pair contributed its sampling difference to the rect permanently, so the selection crept away from the pointer.

The comment above the global monitor already required these handlers to be idempotent — the delta accumulation quietly broke that contract. The translation is now recomputed absolutely from anchors captured when the move begins, so a repeated, stale or out-of-order sample lands on the same rect and nothing accumulates.

### Size and origin changing on release

`endManualSelection` assigned the mouse-up point straight to the current point with no knowledge of move mode, so releasing the mouse while Space was held moved that corner while the start point kept its translation. Because the mouse-up point is its own pointer sample, the committed rect came out a different size *and* origin than the drawn one every time. Drag updates and the final mouse-up now route through one shared transform.

Anchoring on the selection's current point rather than a fresh pointer read makes both seams exactly zero: pressing Space neither moves nor resizes the selection, and releasing it does not snap.

### Per-pointer-event work in the same hot path

- `requestDisplayActivationForManualSelection` ran `NSScreen.screens` plus a per-screen `deviceDescription` build on every event. It is a no-op once every display has a backdrop or live-fallback, and both sets only grow within a session, so it now latches off. Cleared on a new drag, a screen-configuration change and session teardown.
- `renderManualSelectionIfNeeded` fanned every event out to all pooled windows, including displays holding neither the pointer nor the rect. Now gated, with a term that guarantees the pass which clears a display still runs.

### Testing

The move geometry and the render gate are pure and unit-tested (8 tests), including idempotence under stale, duplicate and interleaved samples, and the clear-exactly-once behaviour of the gate. Verified by hand on a single display.

The render gate has **not** been verified on real multi-display hardware — it is covered at the logic level only. Happy to split that part out if you would rather take the correctness fix on its own.

### DIFF
--- a/Snapzy/Services/Capture/AreaSelectionMoveLogic.swift
+++ b/Snapzy/Services/Capture/AreaSelectionMoveLogic.swift
@@ -2,28 +2,19 @@
 //  AreaSelectionMoveLogic.swift
 //  Snapzy
 //
-//  Pure geometry for space-to-move, extracted so it can be tested without a live overlay session.
-//
 
 import CoreGraphics
 
 /// Anchors captured at the instant a space-to-move begins.
 struct AreaSelectionMoveAnchors: Equatable {
-  /// The selection's start point when the move began.
   var originStart: CGPoint
-  /// The point the translation is measured from — the selection's current point when the move
-  /// began, not a fresh pointer read. Anchoring here makes both seams zero: the selection neither
-  /// resizes when Space goes down nor snaps when it comes back up.
+  /// The selection's current point when the move began, not a fresh pointer read.
   var anchor: CGPoint
 }
 
 enum AreaSelectionMoveLogic {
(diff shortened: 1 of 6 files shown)


duongductrong (approved):
lgtm! thanks @vxirau

PR #505, opened 2026-08-17
feat: show progress and outcome for OCR capture

vxirau:
An OCR capture reports nothing between mouse-up and the text landing on the clipboard except the menu-bar spinner, so a slow recognition reads as a freeze.

### Progress

A toast now appears synchronously on mouse-up, before the first async hop. It is shown only after the mouse-up snapshots are secured, so the panel can never end up in the captured pixels.

One label covers the whole operation. Relabelling once the image is in hand sounds better but is not: the normal path only crops the snapshot already taken at mouse-up, which finishes well inside the toast's own fade-in, so the first label never became visible.

### Outcome

The toast resolves into a terminal state rather than being dismissed. `OCRResultNotifier` prefers a native notification and deliberately shows no toast when one is delivered, and it reports from a detached task that resolves after the capture scope exits - so an unconditional dismiss always won the race and the spinner vanished into nothing.

Each exit now lands on a visible outcome:

| Outcome | Toast |
| --- | --- |
| Text copied | success, "Copied to clipboard" |
| No text found | warning |
| QR-only, unsupported payload | warning |
| Capture or recognition threw | error, with the description |

Notification behaviour is unchanged - this only adds the in-app confirmation for users who miss it, or who have Focus enabled and never receive it.

### Notes

Reuses the existing `AppToastManager` spinner presentation already used by the Annotate editor's OCR, so no new UI. Adds one string, `ocr.capturing-content`, localized across all ten supported languages.

### DIFF
--- a/Snapzy/Features/Capture/CaptureViewModel.swift
+++ b/Snapzy/Features/Capture/CaptureViewModel.swift
@@ -2413,13 +2413,18 @@ final class ScreenCaptureViewModel: ObservableObject, KeyboardShortcutDelegate {
           // Secure pixels first: immediately after the snapshot is taken, dismiss the overlay.
           AreaSelectionController.shared.cancelSelection()
 
-          let progressToast = AppToastManager.shared.show(
-            message: L10n.OCR.capturingContent,
-            style: .info,
-            duration: nil,
-            variant: .compact,
-            iconMode: .spinner
-          )
+          // The menu-bar spinner is the primary progress surface, but `setProcessing` is a no-op
+          // without a status item, so it reports nothing when the icon is hidden. Fall back to a
+          // toast only in that case, never alongside it.
+          let progressToast = AppStatusBarController.shared.isMenuBarIconVisible
+            ? nil
+            : AppToastManager.shared.show(
+              message: L10n.OCR.capturingContent,
+              style: .info,
+              duration: nil,
+              variant: .compact,
+              iconMode: .spinner
+            )
 
           Task { @MainActor in
             var progressToastResolved = false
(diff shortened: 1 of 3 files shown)


duongductrong:
Hi @vxirau, love to see your PR again! and thanks for the PR and for taking the time on this! 🙏

One thing though — OCR progress and outcome feedback is already implemented on the app side: while the user is running OCR, the state is surfaced through Snapzy's menubar icon. So this change would end up duplicating that behavior with a second progress indicator.

Could we drop the progress/outcome UI from this PR and rely on the existing menubar feedback instead? If you feel the menubar indication isn't clear enough in some cases, happy to discuss improving that path directly rather than adding a parallel one.

vxirau:
Thanks for the review!

Honestly, I had no idea the menubar icon showed progress during OCR. I have "Show menu bar icon" turned off in Preferences > General, so I've never actually seen that feedback, which is why I assumed there was nothing there at all. My bad for not checking that first.

But digging into it, I think that's the real gap here. `setProcessing(_:)` bails out early when there's no status item.

With `showMenuBarIcon` off, `removeStatusItem()` sets `statusItem = nil`, so that guard fails and the call ends up doing nothing. So for anyone running with the icon hidden, an OCR capture gives no feedback at all. That's the case I hit: between releasing the mouse and the text landing on the clipboard, there's no sign the app is doing anything.

So instead of a second indicator, what about only showing the toast when the menubar icon is hidden? Something like:

```swift
let progressToast = AppStatusBarController.shared.isMenuBarIconVisible
  ? nil
  : AppToastManager.shared.show(...)
```

If the icon is visible, nothing changes at all and the spinner stays the only surface. If it's hidden, the toast covers it instead. They'd never both show up, so no duplication.

Happy to adjust if you'd rather solve it another way, or to scope this down to just the progress part and drop the outcome states. Let me know what you think!

duongductrong (approved):
Thanks @vxirau for adjusting the implementation! Checking `AppStatusBarController.shared.isMenuBarIconVisible` makes complete sense and nicely solves the hidden menu bar icon case without duplicating feedback indicators. The code and localization changes look very clean.

Approving this PR now — I won't merge it immediately just yet as we are organizing the upcoming release cycle, but will merge it during the next release batch. Thanks again for your great contribution! 🙏

vxirau:
No problem @duongductrong ! ;)

PR #510, opened 2026-08-19
fix: render canvas shadow in screenshot exports

RaviMaru20:
## Summary

Canvas/preset drop shadows are visible in Annotate but were missing from exported PNGs.

The shadow intensity was preserved correctly through presets and render snapshots. The loss occurred in both normal flat export paths: they drew the canvas background, clipped to the screenshot shape, and drew the source image without first compositing the screenshot shadow.

This change:

- shares the existing Annotate preview shadow radius and offset with export rendering
- composites a Core Graphics shadow from the rounded screenshot silhouette before source clipping
- keeps the shadow outside the screenshot bounds so source pixels remain untouched at native resolution
- covers manual Annotate export and automatically applied preset export
- preserves existing output when shadow intensity is zero
- does not rely on ScreenCaptureKit native window shadows

## Tests

Added pixel-level regression coverage for:

- rounded-corner shadow over a solid background
- zero shadow intensity with unchanged background pixels
- square shadow over a gradient background
- 2x source pixel preservation with no softened edge pixels

Validation:

- AnnotateExporterTests: 12/12 passed
- PostCaptureActionHandlerTests: 20/20 passed
- unsigned Debug build passed
- git diff --check passed

The exact documented signed build/test commands cannot run on this machine because the upstream development signing certificate for team XMHV5GH2Z7 is unavailable. An unsigned full-suite run also exposed unrelated environment/parallel instability, including a multi-display coordinate-space assumption and isolated tests that pass independently.

## Manual verification

The Release-configuration app was installed in /Applications with native window shadow disabled. The user confirmed macOS permissions are detected and the canvas/preset shadow appears in the running app export workflow.

### DIFF
--- a/Snapzy/Features/Annotate/AnnotateManager.swift
+++ b/Snapzy/Features/Annotate/AnnotateManager.swift
@@ -9,7 +9,10 @@ import AppKit
 import Foundation
 
 enum AnnotateCanvasDefaults {
-  static let cornerRadius: CGFloat = 0
+  nonisolated static let cornerRadius: CGFloat = 0
+  nonisolated static let shadowRadius: CGFloat = 15
+  nonisolated static let shadowOffsetX: CGFloat = 0
+  nonisolated static let shadowOffsetY: CGFloat = 8
 }
 
 /// In-memory annotation session data for re-editing annotations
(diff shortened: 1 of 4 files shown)


duongductrong:
Hi @RaviMaru20, Thank you for your contribution, lgtm!

PR #519, opened 2026-08-22
feat: improve area selection dimension label

### LINKED ISSUE #431: [Feature]: Show Image dimensions when capturing area
### Problem or Motivation

When drawing an area to capture a screenshot, e.g. for a documentation, images need to have the same dimensions so it looks good

### Proposed Solution

While drawing the area, in the overlay there should be clearly visible what dimensions (height, width) the selected area has. Yes, in the bottom right corner there is something like that, but this is pretty small and hard to read. 

### Alternatives Considered

_No response_

### Feature Area

Screenshots

### Mockups / References

_No response_

### Additional Context

_No response_

motoish:
## Summary

Fixes #431.

Improves the area selection dimension label so the selected width and height remain easy to read while dragging.

- Display dimensions in a single-line `width × height` format
- Use a larger monospaced font with a dark, rounded background
- Position the label near the bottom-right corner of the selection
- Keep the label inside the overlay when the selection approaches screen edges
- Show the label only on the overlay containing the active pointer
- Preserve the dimension label across layout and other passive updates
- Restore the existing coordinate indicator style after the selection ends

## Testing

- Added coverage for dimension formatting, presentation, positioning, and edge fallback
- Added a hosted-window regression test for layout updates
- Verified coordinate indicator style restoration
- Verified existing passthrough cursor and stale-location behavior

### Before

1affc5290db8b34759b4ef764330b7c4

### After

ebd01ea1fbe2158b541ff66edfc8c0b7

### DIFF
--- a/Snapzy/Services/Capture/AreaSelectionWindow.swift
+++ b/Snapzy/Services/Capture/AreaSelectionWindow.swift
@@ -3455,7 +3455,10 @@ final class AreaSelectionOverlayView: NSView {
 
   private func refreshDimensionIndicatorAfterPassiveUpdate() {
     guard let localRect = lastVisibleSelectionLocalRect else { return }
-    guard showsDimensionIndicator(forLocalPointer: currentLocalMousePoint()) else {
+    // During an active drag the tap/window path keeps `currentMousePosition` fresh;
+    // `currentLocalMousePoint()` reads `NSEvent.mouseLocation`, which unit/CI hosts
+    // do not update when tests synthesize mouse-moved events.
+    guard showsDimensionIndicator(forLocalPointer: currentMousePosition) else {
       hideSizeIndicator()
       return
     }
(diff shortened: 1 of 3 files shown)


duongductrong:
Hi @motoish,

Love to see your PR again. Thanks for the suggestion. I'd prefer not to switch coordinators just because a few previous versions of Snapzy did it. I decided to stick with the current version, staying as close to the macOS built-in behavior as possible, so everything feels familiar and friendly to users.

I think we can make this a theme coordinator instead, keep the native style as the default, and add a few options for users to select from if needed. (please make sure if implementing theme coordinator, it must be not affect to performance of the flow)

Also worth noting: this ticket was raised by just one user, so forcing a change for everyone based on a single report doesn't feel right.

motoish:
> Hi @motoish,
> 
> Love to see your PR again. Thanks for the suggestion. I'd prefer not to switch coordinators just because a few previous versions of Snapzy did it. I decided to stick with the current version, staying as close to the macOS built-in behavior as possible, so everything feels familiar and friendly to users.
> 
> I think we can make this a theme coordinator instead, keep the native style as the default, and add a few options for users to select from if needed. (please make sure if implementing theme coordinator, it must be not affect to performance of the flow)
> 
> Also worth noting: this ticket was raised by just one user, so forcing a change for everyone based on a single report doesn't feel right.

@duongductrong 

Understood. Thanks for clarifying.

You’d prefer to keep the current macOS-like behavior as the default. I’ll close this PR for now.
Happy to revisit later if you want an optional theme approach.

PR #539, opened 2026-09-02
fix: capture remapped and consumed keystrokes

qeude:
## What was wrong
The keystroke overlay could miss shortcuts when another app remapped modifier keys or consumed keyboard events. In other cases, Snapzy displayed the key but dropped the rewritten modifiers.

For example, a remapped Control + Option + Command shortcut could appear as only `1` or `→`.

Restarting a recording could also leave overlays visible on screen but absent from the exported video: the coordinator retained the overlay windows, while preparing the new recording cleared their capture-filter exceptions.

## What changed
Snapzy observes key-down events at the start of the session event stream and always forwards the original event unchanged. It keeps the normalized modifier state reported after remapping so the overlay can display the complete shortcut.

Normal start, microphone retry, and restart use the same overlay setup path. That path now awaits registration of all effect overlays with the new recording stream, including retained windows. Annotation, click-highlight, and keystroke overlays are covered without recreating them or clearing annotations.

## How to test
1. Enable Show Keystrokes and click highlighting.
2. Start a recording and use remapped Control + Option + Command shortcuts with a number and an arrow key.
3. Check that modifiers and keys are shown correctly.
4. Add an annotation, restart the recording, then show more shortcuts and click effects.
5. Stop and inspect the exported video, not only the live overlay. All enabled effects should still appear.
6. Repeat the shortcut check after continuing without microphone access.

## Validation
- Keystroke event regression tests pass on macOS with Xcode 26.5.
- Replayed the prepare/start/cancel/prepare/start sequence against the actual coordinator and ScreenCaptureKit stream. All three retained overlay IDs were registered in both streams.
- Inspected a frame from the restarted recording: the keystroke badge is present in the exported video.
- Restart verification used a local debugger-driven recording, not a new automated unit test.
- The full suite was not rerun for this follow-up. Earlier full-suite runs encountered existing environment-sensitive failures around mouse position, OCR teardown, area selection, and pasteboard state.


Snapzy/Services/Capture/KeystrokeMonitorService.swift
```diff
@@ -12,62 +12,129 @@ import Foundation
...
+    observedModifierFlags = flags.intersection(.deviceIndependentFlagsMask)
+  }
+
   private func handleKeyDown(_ event: NSEvent) {
     // Ignore key repeats to avoid spamming
     guard !event.isARepeat else { return }
 
-    let flags = event.modifierFlags.intersection(.deviceIndependentFlagsMask)
+    let flags = event.modifierFlags
+      .union(observedModifierFlags)
```
duongductrong, line 134: Automated AI Code Review — Request Changes

P1 — `observedModifierFlags` is sticky: the tap only listens for `keyDown`, while this state is updated by AppKit `flagsChanged` monitors. If a modifier-release event is consumed by a remapper or is not delivered to the monitor, the old flags remain and are `union`ed into every subsequent keyDown. For example, after pressing `⌘` and losing the release event, typing `password` could emit `⌘ P`, `⌘ A`, ... into the recording. This is a correctness issue and may unintentionally expose keystrokes in the video.

Recommendation: track `flagsChanged` in the same CGEvent tap/state path and reset the state reliably (or add an expiry/session reset). Please also add a regression test for a lost modifier-release followed by an unmodified key.

Snapzy/Features/Recording/RecordingCoordinator.swift
```diff
@@ -744,14 +745,7 @@ final class RecordingCoordinator: ObservableObject {
...
 
-        // Setup annotation overlay (must be after recording starts so window exists)
-        setupAnnotationOverlay(for: rect)
-
-        // Setup click highlight overlay (must be after recording starts)
-        setupClickHighlightOverlay(for: rect)
-
-        // Setup keystroke overlay (must be after recording starts)
-        setupKeystrokeOverlay(for: rect)
+        await setupRecordingOverlays(for: rect)
```
duongductrong: Automated AI Code Review — Request Changes

P1 — `setupRecordingOverlays` is awaited after the recorder has transitioned to `.recording`, and the helper may perform up to three asynchronous `SCShareableContent`/`updateContentFilter` operations. A Stop shortcut can therefore interleave during this window: `cleanup()` closes and nils the toolbar, then the start task resumes and still calls `window.showRecordingStatusBar(...)` on the next line using the stale local reference. This can cause the status bar to reappear after recording has stopped or leave UI/callback state stale.

Recommendation: manage/cancel the start task or use a generation token. After each `await`, verify `isActive`, `recorder.state == .recording`, and `toolbarWindow === window` before mutating the UI. Also consider batching overlay registration so the content filter is updated only once.

duongductrong (changes requested):
Automated AI Code Review — Request Changes

PR #539 is directionally correct: it observes keystrokes before downstream event taps and restores recording overlays after a restart. However, two P1 correctness/privacy and lifecycle risks remain, as detailed in the inline comments below. Please address both issues and add regression coverage before merging.

Snapzy/Services/Capture/KeystrokeMonitorService.swift
qeude: Addressed in `693de0ed`. The keyboard tap now observes `flagsChanged` events and clears stale modifier state, preventing consumed modifier releases from leaking into subsequent plain keystrokes. Added a regression test for this case

Snapzy/Features/Recording/RecordingCoordinator.swift
qeude: Addressed in `693de0ed`. Recording startup now uses a generation check after each `await`; Stop, Cancel, Delete, and Restart invalidate stale startup work before it can update the UI. Added lifecycle tests for interrupted and repeated overlay registration.

PR #546, opened 2026-09-05
Fix/draw over annotations

### LINKED ISSUE #535: [Feature]: Layering should be more accommodating
### Problem or Motivation

I find that if I add one annotation object, then select another tool, I cannot click in the vicinity or on the existing annotation to add anything using my new tool. Example: I cannot START drag to create an arrow FROM a box annotation I already created to some other part of the canvas. Or I cannot click to place a test box over a box I already drew to surround something important.

I end up having to move things around or create them somewhere else and then drag them into place.

For times when you want to move something that is part of stacked annotations, maybe we could click the first item easily but use a hot key and click to instead select the next object below, iteratively?

### Proposed Solution

Allow new annotations to originate on top of previous annotations without changing to the select mouse/cursor.

For times when you want to move something that is part of stacked annotations, maybe we could click the first item easily but use a hot key and click to instead select the next object below, iteratively?

### Alternatives Considered

_No response_

### Feature Area

Annotation / Editing

### Mockups / References

_No response_

### Additional Context

_No response_

tukuyomil032:
## Problem

When a drawing tool was active, `DrawingCanvasNSView.mouseDown` hit-tested existing annotations before reaching the new-drawing path. Starting a drag inside an existing rectangle, arrow, or text annotation could therefore move the existing item instead of creating the new annotation the user intended.

Fixes #535

## Solution

- Treat existing annotations as canvas content whenever a drawing tool other than Selection is active.
- Allow Rectangle, Filled Rectangle, Oval, Arrow, Line, Text, Highlighter, Blur, Spotlight, Counter, Watermark, and Pencil to start drawing on top of existing annotations.
- Keep Selection responsible for selecting and moving existing annotations.
- Preserve resize-handle priority, Crop handling, and double-click text editing.
- Preserve annotation z-order and persistence behavior.
- Add regression coverage for drawing over existing annotations, Selection movement, resize-handle priority, all annotation tools, and newly-created annotation selection.

## Scope

This PR covers creating new annotations over existing annotations.

Selecting lower layers in an overlapping stack with a modifier key remains out of scope and can be handled in a follow-up PR after deciding the modifier key, cycling order, wraparound behavior, reset conditions, and hit areas for lines and arrows.

## Validation

All validation tests passed successfully.

## Manual QA

- [x] Created a new rectangle inside an existing rectangle.
- [x] Started an arrow drag from inside an existing rectangle.
- [x] Moved an existing annotation with the Selection tool.
- [x] Resized an existing annotation using its resize handles.
- [x] Added text on top of an existing shape.
- [x] Confirmed that using a drawing tool does not move the annotation underneath it.
- [x] Confirmed that the new annotation becomes selected after creation.
- [x] Attached a screen recording showing the real-app interaction.
- [x] Attached a screenshot showing the resulting nested annotations.

## Manual QA Evidence

The attached recording and screenshot show the real-app verification of drawing, moving, resizing, and adding text around existing annotations.
Snapzy-fix-draw-over-annotation

https://github.com/user-attachments/assets/08f9e7c9-f9b8-49d2-bd0e-c86f038bcb69

## Notes

- Updated `docs/ANNOTATE.md` with the new drawing behavior.
- No persistence format or z-order format was changed.


duongductrong (approved):
LGTM

PR #584, opened 2026-09-18
fix: Restore pinned window dragging on macOS 27

### LINKED ISSUE #544: [Bug]: Pinned screenshot is not movable on MacOS 27
### Description

when the screenshot is pinned, it can't be dragged with the cursor on any other window management app

### Steps to Reproduce

1. take a screenshot
2. pin it
3. try moving

### Expected Behavior

should be 

### Actual Behavior

scrolling for scaling works, dragging doesn't

### Screenshots / Recordings

_No response_

### Snapzy Version

1.31.0 (173)

### macOS Version

### Installation Method

DMG download

### Additional Context

27.0 Beta (26A5425a) - Dev Beta 8

### LINKED ISSUE #575: [Bug]: macOS 27 使用截图钉在屏幕功能时，无法正常拖拽移动图片
### Description

更新macOS27后出现，钉在屏幕后，可以正常缩放、锁定，但无法拖动位置。
最初未意识到是系统版本问题，尝试使用官方提供的脚本完全卸载并安装历史版本后仍无法拖动
最后用朋友的低系统版本mac测试可以正常拖动

### Steps to Reproduce

**

### Expected Behavior

**

### Actual Behavior

**

### Screenshots / Recordings

_No response_

### Snapzy Version

1.32.0 and 1.30.1

### macOS Version

### Installation Method

None

### Additional Context

_No response_

archibald-nice:
Fixes #575. Fixes #544.

## Problem

On macOS 27, pinned screenshots can no longer be repositioned by dragging the image area. Zooming and locking still work; only moving is broken.

## Root cause

Pin windows rely on `isMovableByWindowBackground = true`, and on macOS 27 AppKit no longer initiates background window drags when the content view is a SwiftUI `NSHostingView`. I verified this on macOS 27.0 (build 26A428) with a minimal interactive repro:

| Content / approach | Drag result |
| --- | --- |
| `NSHostingView` (current implementation) | does not move |
| `NSHostingView` + `mouseDownCanMoveWindow = true` override | does not move |
| Plain AppKit `NSView` content | moves normally |
| `NSHostingView` + manual drag via `sendEvent` | moves normally |

So the regression is specific to the SwiftUI hosting-view path in AppKit; the documented `mouseDownCanMoveWindow` contract does not restore it.

## Fix

`QuickAccessPinWindow` layers a native-first drag decider (`PinWindowDragDecider`) on top of the untouched `isMovableByWindowBackground` path — no version gating:

- Every eligible drag starts in a `.tracking` state and events keep flowing to AppKit untouched, so on systems where native background dragging works (macOS 13–26, or a future AppKit fix) behavior is exactly as before.
- Only when the frame demonstrably stays put past a 4pt movement threshold does the manual path take ownership and follow the cursor via `setFrameOrigin`. Frame movement is compared with a small tolerance so point snapping during a live drag cannot false-trigger the backoff.
- If the system moves the frame at any point during the gesture (e.g. native dragging is restored in a future update), ownership is ceded back immediately — exactly one mode ever owns a gesture.
- Interactive chrome (close, zoom, lock, and the bottom file-drag handle) is excluded from background dragging via `PinWindowDragExclusionView` markers, with geometry converted into the hosting view's flipped coordinate system, so a press on a control never moves the window and the control keeps its click.
- State hygiene: a fresh mouse-down always resets, a missed mouse-up is detected via `pressedMouseButtons`, resigning key status ends the gesture, and a manual drag swallows the trailing mouse-up so the click underneath does not fire.

All other event flows (scroll-wheel zoom, magnification gesture, escape-to-close, lock passthrough) are untouched.

## Testing

- `xcodebuild test -scheme Snapzy -destination 'platform=macOS' -only-testing:SnapzyTests/QuickAccessPinWindowDragDeciderTests CODE_SIGNING_ALLOWED=NO` — 9/9 tests pass, covering manual ownership, native cession, point-snapping tolerance, and stale-state resets.
- macOS 27.0 (build 26A428), Apple Silicon, verified on device:
  - Screenshot → pin → drag: pinned window follows the cursor for the whole gesture.
  - Press-and-drag on the close / zoom / lock buttons: window stays put, no action fires on release; plain clicks on the buttons work normally.
  - Bottom drag-to-export handle, scroll-wheel zoom, lock/unlock all work as before.
  - App diagnostic log shows zero WRN/ERR entries during the session; no crash reports.

Demo (recorded on macOS 27 with this branch):

<video src="https://github.com/archibald-nice/Snapzy/raw/fix/pin-window-drag-macos-27/docs/attachments/pin-drag-macos-27-demo.mp4"></video>

The demo video file is only there to embed in this PR description — happy to drop it from the merge if you'd rather not carry it in the repo.


duongductrong (changes requested):
Hi @archibald-nice, thanks for your contribution, I have a requesting changes before merge, please help me review it.

The macOS versions below 27 path is safe: the version gate leaves native AppKit event delivery and isMovableByWindowBackground behavior intact.

For macOS 27+, please make this a true native-first fallback rather than a version-only manual-drag policy:

1. The >= 27 ProcessInfo gate selects manual dragging unconditionally. It does not check API availability or whether native background dragging actually works. The APIs used here are already supported by the macOS 13 deployment target, so an availability check alone will not detect this regression. Forward an eligible background drag to AppKit first and detect whether the window frame moved; enable manual dragging for that gesture only when native dragging did not move it. Ensure exactly one mode owns the gesture.

2. Manual dragging currently excludes only the file-drag handle. It must also exclude close, lock, zoom, and any future interactive controls. A drag that starts on a control can move the panel and still trigger its action on mouse-up.

3. Reset manual drag state on a fresh eligible mouse-down and on cancellation, deactivation, or a missed mouse-up. Otherwise a stale offset can make the next drag jump or fail.

Please add regression coverage for the version policy and manually verify macOS 13/latest 26 remain native, plus macOS 27 native-working and native-broken paths.

archibald-nice:
Thanks for the thorough review — all three points are addressed in the latest push (`a81d665`):

**1. Version gating → native-first with runtime detection.** The gate is gone. Every eligible drag now starts in a `.tracking` state where events flow to AppKit completely untouched, and the decider watches the window frame: it only switches to the manual path when the frame has stayed put past a 4pt movement threshold, and it cedes back to `.native` the moment the system moves the frame on its own. Since a live drag can be subject to point snapping, the frame comparison uses a small tolerance rather than exact equality. macOS 13–26 therefore keep the pristine AppKit behavior, and a future AppKit fix takes precedence automatically — no OS checks anywhere.

**2. All interactive chrome excluded.** Close, zoom, lock, and the bottom file-drag handle are all wrapped in `PinWindowDragExclusionView` markers now (`interactiveRegion(_:)`), not just the file handle. One real-world subtlety worth knowing: the marker is realized as a *sibling* behind its control (SwiftUI `.background`), so ancestor hit-testing can't see it — the window walks the content view's subtree and tests marker bounds directly. The hosting view is also flipped, so all geometry is converted via `convert(_:from:)` / `convert(_:to:)` instead of mixing window-base and view coordinates.

**3. State hygiene.** A fresh mouse-down always resets before evaluating; a missed mouse-up is detected each drag event via `NSEvent.pressedMouseButtons`; `resignKey()` ends the gesture; and once the manual path owns a drag it swallows the trailing mouse-up, so the click underneath never fires (matching native window-drag semantics — press-drag-release on a button neither moves the window nor triggers the action).

Also added `QuickAccessPinWindowDragDeciderTests` — 9 cases covering manual ownership, native cession (including mid-gesture), snapping tolerance, and stale-state resets. Verified on device on macOS 27.0 (build 26A428): full-gesture tracking, buttons stay put under press-drag while plain clicks still work, export handle and zoom untouched, zero warnings in the diagnostic log.

The PR description has been rewritten to match the new implementation. Ready for another look! 🙏

duongductrong (approved):
lgtm

archibald-nice:
Merged — thank you! This one was a fun deep dive into AppKit event flow. The native-first design ended up much cleaner thanks to your review feedback.

Also, it was almost midnight your time when you merged this — get some rest. 😄

SoulXM:
The BUG has been fixed. Thank you

PR #586, opened 2026-09-19
feat: Quick Access - Add Sound Effects Toggle

rvanhorn:
Hello!

Thanks for creating this app, was looking for something to replace Clean Shot X. 

I noticed the Quick Access panel has its own sound effect with no way to disable it so I added an option for it. 

## Summary

- Add a Sound Effects toggle under Settings → Quick Access.
- Allow users to silence the Quick Access overlay and card-action sounds without disabling the overlay.
- Keep capture success and failure sounds under the existing global sound preference.
- Add `quick_access.play_sounds` to TOML configuration import and export.
- Update localization, documentation, and focused tests.

## Testing

- Ran `QuickAccessCoreTests`.
- Ran `SnapzyConfigurationServiceTests`.
- Ran `SnapzyConfigurationImporterTests`.
- Verified the localization catalogs.
- Built, ad-hoc signed, and installed.
- Confirmed that the new toggle appears, persists after reopening Settings, and disables Quick Access sounds.
- The full test suite was run.


duongductrong:
Thanks for working on this and for addressing this request.

I understand the distinction from the existing Settings → General → Play sounds preference:

- General → Play sounds is the global master switch for all Snapzy sounds.
- Quick Access → Sound Effects provides a scoped control for Quick Access overlay and card-action sounds.

This is a useful addition for users who want to keep other Snapzy sounds enabled while silencing Quick Access sounds.

A couple of follow-ups before approval:

- Please ensure all Quick Access sound paths honor the new preference. Some copy and cloud-upload flows still call SoundManager.play("Pop") directly.
- Please include quick_access.play_sounds in SnapzyConfigurationDefaultDocument so Restore Defaults also resets this preference.

Overall, the direction looks good and matches the requested use case. Thanks again for the contribution!

duongductrong:
One additional note on localization: the two new user-facing strings currently only have English entries, while Snapzy supports 10 locales and the localization guidelines ask us to localize all user-facing copy.

The catalog validator still passes because it checks key coverage, not whether every supported locale has a translation. Please consider adding translations for the remaining locales, or note this explicitly as a follow-up if translations are being handled separately.

This is not a security or performance concern, but otherwise users on non-English locales will see these settings in English.

rvanhorn:
Hello,

Made the requested changes and added the translations, hopefully this covers everything. :)

duongductrong (approved):
lgtm

PR #594, opened 2026-09-20
fix: apply text background color by promoting plain text to label

### LINKED ISSUE #592: [Bug]: Mac OS27 区域截图文字描述 背景色无响应
### Description

Mac OS27 环境下使用 区域截图后，添加文字描述的 背景色无响应

Image

### Steps to Reproduce

1. Capture Markup
2. Add text
3.Setup background Color

### Expected Behavior

Change the background color of text area, and can select the shape of area, just like annotate.

Image

### Actual Behavior

No background color effected, and no shape can be selected

### Screenshots / Recordings

_No response_

### Snapzy Version

1.32.1（190）

### macOS Version

### Installation Method

DMG download

### Additional Context

_No response_

archibald-nice:
Fixes #592

## Problem
Picking a background color for a text annotation (sidebar grid, quick properties bar, or inline area annotate bar) appeared to do nothing, and the inline bar offered no way to switch the text shape.

## Root cause
The background fill only renders when `textPresentation` is `.label`/`.callout` (`AnnotateAnnotationRenderer`, `TextEditOverlay`), and new text annotations default to `.plain`. Every background color control wrote `fillColor` without leaving `.plain`, so the color was stored but never drawn. The shape switcher exists in the editor's quick properties bar but was never added to the inline bar.

## Changes
- `AnnotateState.updateAnnotationProperties`: when a non-clear fill is applied to a `.plain` text annotation, promote `textPresentation` to `.label`. This covers the sidebar, quick bar, inline bar, and multi-select batch edits, since they all funnel through this method.
- `AnnotateState.updateDefaultAnnotationProperties`: same promotion for the text tool's default properties, so a remembered default background also produces visible labels for new text.
- `InlineAreaAnnotateWindow`: add the text presentation switch (plain/label/callout) to the inline properties bar, next to the background color control.

## Semantic note
One behavior changes deliberately: a text annotation switched back to `.plain` keeps its stored fill color (existing design). If the user then changes that fill color, it now promotes to `.label` and becomes visible, instead of silently storing an invisible color. Happy to adjust if you'd prefer a different interaction.

## Testing
- Added 5 tests in `AnnotateTextEditingTests` (promotion, non-plain unchanged, clear fill no-op, quick-binding paths).
- Ran 236 annotate-related unit tests, all passing. Two failures in `AreaSelectionMultiMonitorReconciliationTests` / `AppStatusBarControllerTests` reproduce on clean `master` and are environment-related.
- Manual regression: inline bar, editor sidebar, quick bar, default-style memory, export PNG, undo/redo consistency, shape switching round-trips, other tools (rect/arrow/highlight/blur/counter/watermark), mixed multi-select, session restore.


Snapzy/Features/Annotate/AnnotateState.swift
```diff
@@ -4659,6 +4666,12 @@ final class AnnotateState: ObservableObject {
         properties.fillColor = fillColor
       }
     }
+    if tool == .text,
```
duongductrong, line 4685: [P1] Persist the promoted text presentation with the tool default

This promotion works only for the current in-memory `AnnotateState`.

When no text is selected, choosing a non-clear background calls `updateDefaultAnnotationProperties`, which sets the fill, promotes `.plain` to `.label`, and then calls `persistAnnotationToolProperties()`. However, the private `PersistedAnnotationProperties` model (lines 101-147) does not store `textPresentation`.

Repro:
1. With no text selected, choose yellow as the text background.
2. Create text. The background is visible.
3. Relaunch the app and create text again.
4. The saved yellow fill is restored, but the presentation falls back to `.plain`.
5. `AnnotateAnnotationRenderer.drawText` only paints the background when `textPresentation != .plain`, so the original issue returns.

Please persist `textPresentation` as an optional field for backward compatibility and add a reload/two-state regression test.

The new inline style picker has the same gap: when no text is selected, `setTextPresentation` updates `annotationToolProperties` and returns without calling `persistAnnotationToolProperties()` (lines 5024-5033). Please cover that path as well.

Snapzy/Features/Annotate/AnnotateState.swift
```diff
@@ -3748,6 +3748,13 @@ final class AnnotateState: ObservableObject {
     }
     if let fillColor = colorUpdate.fillColor {
       annotations[index].properties.fillColor = fillColor
+      // The background fill only renders for label/callout presentations, so a
+      // colored fill on plain text would stay invisible. Promote it to a label.
+      if case .text = annotations[index].type,
```
duongductrong, line 3769: [P2] Include the promotion in the early-return change check

The new promotion block is after `annotationPropertiesWillChange`, so it is skipped when the selected fill is already equal to the current fill.

This is reachable through the behavior in this PR:
1. Set a yellow background; the text becomes `.label`.
2. Change the text style to `.plain`; the fill remains yellow.
3. Select yellow again in the background picker.
4. `annotationPropertiesWillChange` returns false because the fill value did not change, so the promotion at this line is never executed. The swatch remains yellow, but the text stays `.plain` and the renderer hides the background.

Please include the `.plain -> .label` promotion in the change predicate (or normalize it before the early return) and add a regression test for `plain + existing yellow fill + selecting yellow again`.

duongductrong (commented):
Thanks for the focused fix for #592. I verified commit `dc8380f2`: the added AnnotateTextEditingTests pass, and the PR build/SwiftFormat checks are green.

Before merging, please address the two behavior gaps in the inline comments. The current tests cover the initial `clear -> non-clear` transition, but not persistence across relaunch or the `plain + existing non-clear fill -> same color` path.

archibald-nice:
Thanks for the thorough review! Both gaps are addressed in `543e523a`:

**P1 — persistence:** `PersistedAnnotationProperties` now stores `textPresentation` as an optional field, so the promoted `.label` default survives relaunch alongside the restored fill. The field decodes to `.plain` when absent, keeping old payloads compatible. `setTextPresentation` also writes through to persisted tool defaults when no text is selected.

**P2 — same-fill re-selection:** `updateAnnotationProperties` now treats a non-clear fill on plain text as a meaningful change even when the fill value is unchanged, so re-picking the current color after switching back to `.plain` still applies the promotion.

Regression tests added for all three paths (promotion across state reload, inline presentation persistence, and the same-fill re-selection case). `AnnotateTextEditingTests` is 29/29 green locally, and I also verified the three scenarios manually on a debug build: background survives relaunch, the inline presentation picker survives relaunch, and re-selecting the same fill after switching to plain brings the background back.

duongductrong (approved):
lgtm

